PR #11, opened 2018-07-11
U14069: Homepage Markup

lukemoderwell:


Gemfile
```diff
@@ -5,7 +5,7 @@ gem "jekyll", "~> 3.7.3"
 group :jekyll_plugins do
   gem 'jekyll-assets'
   gem 'crds-styles', git: 'https://github.com/crdschurch/crds-styles.git', branch: 'development'
-  gem "jekyll-contentful", git: 'https://github.com/crdschurch/jekyll-contentful.git', branch: 'development'
+  gem "jekyll-contentful", git: 'https://github.com/crdschurch/jekyll-contentful.git', branch: 'master'
```
tcmacdonald: We need to specify the version number (eg `~> 0.0.7`) here since `master` will soon contain `v1.0` of this library, which contains some breaking changes.

_config.yml
```diff
@@ -68,3 +74,52 @@ contentful:
...
+          end_date: ends_at
+          youtube_url: youtube_url
+          apple_podcasts_url: apple_podcasts_url
+          google_play_url: google_play_url
+          year: "{{ starts_at | date: '%Y' }}"
+          videos: videos/id
+        other:
+          layout: series
+          draft: false
+    messages:
```
tcmacdonald: Now that we're pulling in series and messages into this project, we have a new urgency to limit the volume of content coming back from Contentful. For example, in production today we have 1616 message records which will dramatically lengthen our build times. 

I'd suggest that we fix this issue in jekyll-contentful (via predecessor or blocking user-story) by providing some sort of mechanism to limit the number of records that are pulled back. We really only need the most recent series/messages content right?

lukemoderwell: I think today it would just be the current series and last message.

zarahzachz:
@lukemoderwell The jumbotron isn't full-width in this design. Is it supposed to be?

screen shot 2018-07-18 at 11 45 47 am

PR #12, opened 2018-07-12
US14086 ENV Specific Redirects

tcmacdonald:


lib/crds/redirects.rb
```diff
@@ -0,0 +1,58 @@
+require_relative '../utils/colorized_string'
+require 'csv'
+
```
mjvezzani: Thinking as a developer who would be new to this project and not be familiar Ruby, I'm thinking of the ways that cognitive overhead can be decreased. This file is called `redirects.rb`, but it doesn't contain the actual redirects. Instead it is responsible for writing the redirects. So would it be better names `redirects_writer` or something like that?

lib/crds/redirects.rb
```diff
@@ -0,0 +1,58 @@
...
+        path, dest, status = row
+        dest = replace(dest)
+
+        printf ColorizedString.new("\s\s%-#{tabs[0]}s %-#{tabs[1]}s %s\n").send(:yellow), path, dest, status unless silence
+        @output.puts("#{path}\t#{dest}\t#{status}")
+        i += 1
+      end
+
+      log "#{i} rows written to _redirects file", :green unless silence
+      @output.close
```
mjvezzani: This method feels a bit long. Any way you could leverage private methods to extract the different blocks of responsibility and make it more readable for the next developer who has to touch this file?

lib/crds/redirect_writer.rb
```diff
@@ -0,0 +1,66 @@
...
+
+    def initialize(dest_dir=nil)
+      FileUtils.mkdir_p dest_dir unless dest_dir.nil?
+      @output = File.new(File.join(dest_dir || Dir.pwd, '_redirects'), "w")
+      @src = File.join(Dir.pwd, 'redirects.csv')
+      @total = 0
+      @debug = true
+    end
+
+    def write!
```
mjvezzani: 👍

PR #14, opened 2018-07-16
US14119: Implement Location Search

lukemoderwell:


_assets/javascripts/lib/distance-sorter.js
```diff
@@ -0,0 +1,167 @@
...
+    this.locationFinder = new CRDS.LocationFinder();
+  }
+
+  static getLocationsCarousel() {
+    return Object.values(CRDS._instances).find(
+      (instance) => DistanceSorter._isSectionLocationsCarousel(instance)
+    );
+  }
+
+  static _isSectionLocationsCarousel(instance) {
```
lukemoderwell: Nice clean up here from the the original. Reads better

_assets/javascripts/lib/distance-sorter.js
```diff
@@ -0,0 +1,167 @@
...
+    return instance.carousel !== undefined && instance.carousel.id === 'section-locations';
```
lukemoderwell: One small improvement here would be to switch from an `id` to something like a `data-attribute`. I think it could make this a little less brittle...

_assets/javascripts/lib/distance-sorter.js
```diff
@@ -0,0 +1,167 @@
...
+      location: locationObject.location.location,
+      distance: locationObject.distance
+    };
+  }
+
+  _display_sorted_locations_carousel() {
+    this.createDataAttributes();
+    this.appendDistances();
+    this.locationsCarousel.sortBy('distance');
+//  A anywhere card needs to be created in Contenful before
```
lukemoderwell: I actually hard coded this card into the `HTML` - not sure if that was a good approach but it didn't really feel like Anywhere was a true location.

_assets/javascripts/lib/distance-sorter.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  _createLocationDistanceObject(locationObject) {
+    return {
+      location: locationObject.location.location,
+      distance: locationObject.distance
+    };
+  }
+
+  _display_sorted_locations_carousel() {
```
lukemoderwell: what's up with the `_` here?

_assets/javascripts/lib/location-finder.js
```diff
@@ -0,0 +1,22 @@
+/* eslist no-plusplus: 0 */
+/* global CRDS */
+
+window.CRDS = window.CRDS || {};
+
+CRDS.LocationFinder = class LocationFinder {
+  constructor() {
+    this.gatewayAPIEndpoint = 'https://gatewayint.crossroads.net/gateway/api/v1.0.0';
```
lukemoderwell: Still need `env` support here

_assets/stylesheets/_locations.scss
```diff
@@ -17,3 +17,16 @@
     }
   }
 }
+
+.distance.label {
```
lukemoderwell: We could remove most of this css in favor of DDK utility classes.

_assets/javascripts/lib/distance-sorter.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  appendDistances() {
+    for (let i = 0; i < this.cards.length; i += 1) {
+      const locationMatch = this.locationDistances.find(obj => obj.location === this.cards[i].dataset.location);
+      if (locationMatch !== undefined) {
+        this.cards[i].dataset.distance = locationMatch.distance;
+        const span = document.createElement('span');
+        span.classList.add('distance', 'label', 'font-family-base');
```
lukemoderwell: here we can remove `font-family-base` and add `soft-half` and `push-half`. Finished product might look like this:
```
<span class="distance label soft-half push-half">4.9 miles</span>
```

_assets/stylesheets/_locations.scss
lukemoderwell: Specifically lines 23, 24 and 28

PR #18, opened 2018-07-27
Happenings PoC: Promos, Filtering, and more

lukemoderwell:


_config.yml
```diff
@@ -71,6 +71,7 @@ contentful:
           image: image/url
           bg_image: bg_image/url
           map_url: map_url
+          happenings: happenings
```
jasoncallibrity: Didn't we remove this?

lukemoderwell: 👍

PR #19, opened 2018-07-30
Us14211 us14220 de5674 meta description, image, title and display correct page title

agause:


jasoncallibrity:
Just because I know Olga will find it in basically two seconds and to avoid having to work a defect...take a look at index.html

`title: Home | Crossroads` should be `title: Home` now.

_plugins/filters/meta_util.rb
```diff
@@ -0,0 +1,86 @@
...
+    class MetaUtil
+      # Priority:
+      # 1) meta image
+      # 2) page image
+      # 3) page background image
+      # 4) first image found in page content
+      # 5) site variable's image
+      # Don't love it -- some pretty brittle checks based on naming conventions established in media
+      # method for getting the url for the <meta name="image" content="{{ meta_image }}"> tag in the head of a SSG html page
+      def self.get_meta_image(page_meta_image, page_image, page_bg_image, page_content, site_image)
```
tcmacdonald: Checkout [the splat operator](https://medium.freecodecamp.org/rubys-splat-and-double-splat-operators-ceb753329a78). You can use it to create catch-all arguments so you can pass in whatever you want without having predefine all the params you might need... 

```ruby
def self.get_meta_image(*args)
  puts args.inspect     # [ 'https://page-meta-image.jpg', 'https://page-image.jpg', ... ]
end

self.get_meta_image(page_meta_image, page_image, ...)
```

_plugins/filters/meta_util.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        # returns the first image url in the array that has a value
+        image_url = [
+          page_meta_image,
+          page_image,
+          page_bg_image,
+          search_for_first_image_url_in_page_content(page_content), #invoking the regex this way returns the FIRST match
+          site_image
+        ].find{|image_under_review| has_value?(image_under_review)}.to_s.strip
```
tcmacdonald: Checkout  [Array.compact](https://ruby-doc.org/core-2.2.0/Array.html#method-i-compact)

```ruby
image_url = args.compact.shift.to_s.strip
```

_plugins/filters/meta_util.rb
```diff
@@ -0,0 +1,86 @@
...
+        def self.has_value?(to_evaluate)
+          to_evaluate && to_evaluate.to_s.strip.empty? == false
+        end
+
+        # we really only need to account for the scenario when http or https isn't prepended
+        def self.prepend_url_protocol(image_url)
+          url_doesnt_start_with_a_protocol = image_url[/^http[s]?:/].nil?         # asks: does the string NOT start with http: or https:
+          url_starts_with_an_authority_prefix = image_url[/^\/{2}/].nil? == false # asks: does the string start with exactly 2 forward slashes: //
+
+          url_protocol = nil;
```
tcmacdonald: Semicolons are not necessary in Ruby. Stylistically, they are generally omitted.

_plugins/filters/meta_util.rb
```diff
@@ -0,0 +1,86 @@
...
+          if url_doesnt_start_with_a_protocol && url_starts_with_an_authority_prefix # authority prefix being the double forward slash
+            url_protocol = 'https:'
+          end
+
+          "#{url_protocol}#{image_url}"
```
tcmacdonald: Checkout the [URI](https://docs.ruby-lang.org/en/2.1.0/URI.html#module-URI-label-Features) module. You might be able to use it simplify this method.

_plugins/filters/meta_util.rb
agause: @tcmacdonald Force of habit. I'm surprised that's the only one you found. :)

_plugins/filters/meta_util.rb
agause: Excellent, will do. An exact instance where I hoped some module would kill this code.

_plugins/filters/meta_util.rb
agause: @tcmacdonald I'll look to leverage the splat operator wherever possible. Is it applicable where the meta image method is concerned since I need to potentially execute a regex against the 4th argument (page_content)?

agause:
@tcmacdonald Nope, I don't think I've got nearly enough understanding of the processing pipeline to have considered it. But this is the exact type of guidance I was hoping for. Passing "site" around didn't feel awkward, but now I know it should have.

Studying up on the use case for generators will likely make plain why it may be the more natural approach. Thanks man!

_plugins/filters/meta_util.rb
agause: @tcmacdonald Applied this to the get_meta_description method. Thank you!

_plugins/filters/meta_util.rb
agause: @tcmacdonald Couldn't apply this one directly without giving up some functionality, but nevertheless, new tool in my Ruby belt.

_plugins/filters/meta_util.rb
agause: @tcmacdonald Removed.

_plugins/filters/meta_util.rb
agause: @tcmacdonald Leveraged this and shortened the code/complexity here. Perfect!

_plugins/filters/meta_util.rb
```diff
@@ -0,0 +1,86 @@
...
+        end
+
+        #returns the value in the 'src' attribute of the first occurence of an 'img' element
+        def self.search_for_first_image_url_in_page_content(page_content)
+
+          if page_content.nil? || page_content.to_s.strip.empty?
+            nil
+          else
+            matches = /<\s*img .*?src=["']?[\s]*([^'">\s]+)/.match(page_content)
+            matches.nil? ? nil : matches[1]
```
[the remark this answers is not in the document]
agause: @tcmacdonald Wasn't able to apply the dig method here, but definitely got a sense of its usefulness.

agause:
@tcmacdonald pushed up the latest changes with your suggestions, sans the generator solution. I'll circle back to that approach in the future.

_plugins/filters/html_and_meta_util_filter.rb
```diff
@@ -0,0 +1,42 @@
+require_relative '../../lib/utils/meta_util'
+require_relative '../../lib/utils/html_util'
+
+module Filters
```
tcmacdonald: So this is not strictly necessary but you might consider changing the namespace/class name here to something that more closely resembles its place in the file hierarchy... for example:

```ruby
module Filters
  class HtmlAndMetaUtilFilter
    ...
  end
end
```

_plugins/generators/meta_generator.rb
```diff
@@ -0,0 +1,35 @@
+# require_relative '../../lib/utils/html_util'
```
tcmacdonald: Can you delete this file if its no longer needed?

lib/utils/meta_util.rb
```diff
@@ -0,0 +1,70 @@
...
+    class MetaUtil
+      # Priority:
+      # 1) meta image
+      # 2) page image
+      # 3) page background image
+      # 4) page content for 1st image
+      # 5) site variable's image
+      # Don't love it -- some pretty brittle checks based on the naming conventions of present day
+      # method for getting the url for the <meta name="image" content="{{ meta_image }}"> tag in the head of a SSG html page
+      def self.get_meta_image_url(page_meta_image, page_image, page_bg_image, page_content, site_image)
```
tcmacdonald: If you want, we should pair on applying the splat approach to this method sometime.

_plugins/filters/html_and_meta_util_filter.rb
agause: @tcmacdonald Will do that now.

_plugins/generators/meta_generator.rb
agause: @tcmacdonald Yes, mostly just wanted it in there for historical context, should I give the generator another go in the future.

lib/utils/meta_util.rb
agause: @tcmacdonald So this is what I had in the generator after our discussion where I dropped the img url search bit and treated all the content the same...

```ruby
      def self.get_meta_image_url(*image_args)
        image_url = image_args.find{|image_under_review| has_value?(image_under_review)}.to_s.strip
        prepend_url_protocol(image_url)
      end
```
But calling ```search_for_first_image_url_in_page_content``` within the filter again is the reason I reverted back to the distinct method arguments.

PR #21, opened 2018-08-03
US14234: Dynamic location pages

jasoncallibrity:
**Purpose:**
Any 'Location' that exists in Contentful will receive its own location page based on the content that exists in Contentful

**Considerations**
_Uptown_ has a special 'College Students' section hard coded into the template and conditionally displayed
_Andover_ has a Preschool section that displays at the top of the "kids and students" area hard coded 
into the template and conditionally displayed
_Hubspot form_ required the id for the form to be populated, if it is not then a default youtube video is displayed
_Serve_ 'go deeper' card will default to a specific link if none is provided
_Kids Club & student ministry_ will not display if their appropriate fields in Contentful are not populated
_Happenings_ section only displays the Happenings for that site instead of a drop down that has them all


_layouts/location.html
```diff
@@ -1,41 +1,305 @@
...
+              {{ page.address | markdownify | remove: '<p>' | remove: '</p>' }}
+              </br>
+              <a href="{{ page.map_url }}"
+                target="_blank">Get directions</a>
+            </p>
+          </div>
+        </div>
+        <div class="row">
+          <div class="col-sm-12">
+            {% for series in site.series limit: 1 %}
```
lukemoderwell: I think we want `{% assign current_series = site.series | sort: "date" | reverse | first %}` here and then use `current_series` instead of `series`

PR #26, opened 2018-08-10
US14310: Live Stream

lukemoderwell:


_assets/javascripts/components/location-search.js
```diff
@@ -0,0 +1,10 @@
+document.addEventListener("deferred-js-ready", locationSearchInit);
+
+// if app-deferred loads first
+if (window.deferredJSReady) {
+  locationSearchInit();
+}
+
+function locationSearchInit() {
+  new CRDS.DistanceSorter();
+}
```
jasoncallibrity: I think we could probablyl pull this out and make it generic. Seems like this is going to get used on any page with self-specific JS that needs initialized after jquery and other things are loaded. Just a thought

_includes/_footer.html
```diff
@@ -20,8 +20,10 @@ <h5>Resources</h5>
...
-});
-// ]]></script>
\ No newline at end of file
+</div><p></p>
+<script type="text/javascript">// <![CDATA[
+    // $( "#website-feedback" ).click(function() {
+    //     hj('trigger', 'feedback');
+    // });
+    // ]]>
+</script>
```
jasoncallibrity: We should probably discuss this with AG or BM at some point...

PR #34, opened 2018-08-14
- Parses markdownified url from description instead of full page cont…

agause:
…ent when no meta, image or bg_image are specified

- Adds unit test for parsing markdownified url
- Looking explicitly at the page description (instead of the full page content) for the meta image
- Looking explicitly at the page excerpt (instead of the full page content) for the meta description

### DIFF
--- a/lib/utils/meta_util.rb
+++ b/lib/utils/meta_util.rb
@@ -17,7 +17,7 @@
           page_meta_image,
           page_image,
           page_bg_image,
-          search_for_first_image_url_in_page_content(page_content),
+          search_for_first_image_in_content(page_content),
           site_image
         ].find{|image_under_review| has_value?(image_under_review)}.to_s.strip
         prepend_url_protocol(image_url)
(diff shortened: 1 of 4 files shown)


lib/utils/meta_util.rb
```diff
@@ -50,13 +50,31 @@ def self.prepend_url_protocol(image_url)
...
+          search_for_first_markdownified_image_url_in_content(content) ||
+          search_for_first_html_image_url_in_page_content(content)
+        end
+
+        #
+        def self.search_for_first_markdownified_image_url_in_content(content)
+          if content.nil? || content.to_s.strip.empty?
+            nil
+          else
+            matches = /!\[[^\]]*\][\s]*\(([^\)]+)\)/.match(content)
```
lukemoderwell: what is this regex looking for?

lib/utils/meta_util.rb
```diff
@@ -50,13 +50,31 @@ def self.prepend_url_protocol(image_url)
...
+            matches.nil? ? nil : matches[1]
+          end
+        end
+
+#        ![famous-dripping-clocks-dali-painting-persistence-of-memory1](//images.ctfassets.net/p9oq1ve41d7r/53Ilga1G40Sewck6YYOWcI/8ae18b94d0e3aad990c0fff3b440bbda/famous-dripping-clocks-dali-painting-persistence-of-memory1.jpg)
```
lukemoderwell: Guessing this line could be removed

lib/utils/meta_util.rb
```diff
@@ -50,13 +50,31 @@ def self.prepend_url_protocol(image_url)
           uri.to_s
         end
 
+        #
+        def self.search_for_first_image_in_content(content)
+          search_for_first_markdownified_image_url_in_content(content) ||
+          search_for_first_html_image_url_in_page_content(content)
```
lukemoderwell: these two methods seem really similar in what they do. Could you just call `find_first_img` and pass in the context?

PR #44, opened 2018-08-20
DE5780: Live Reminder

lukemoderwell:


_layouts/default.html
```diff
@@ -15,6 +15,13 @@
     {% endif %}
 
     <script>
+    //exist as duplicates due to external dependencies on structure
```
agause: @jasoncallibrity Please expound on the reason(s) these duplicate values exist here in the source for other "humans persons" understanding.

PR #58, opened 2018-08-27
Fix for past weekends series slug

jasoncallibrity:
added logic for capturing series slug from the videos (read as message) they contain

### DIFF
--- a/live.html
+++ b/live.html
@@ -267,14 +267,18 @@ meta_description: Crossroads Live
   <div class="row">
     <div class="card-deck carousel" data-crds-carousel="mobile-scroll">
       <div class="feature-cards card-deck--expanded-layout">
-        {% assign recent_messages = site.messages | sort: 'date' | reverse | slice: 0,4 %} {% for message in recent_messages %}
+        {% assign recent_messages = site.messages | sort: 'date' | reverse | slice: 0,4 %} 
+        {% for message in recent_messages %}
+        {% for series in site.series %}
+          {% assign series_slug = series.slug | where: "videos.slug", message.slug %}
+        {% endfor %}
         <div class="card">
-          <a href="/series/{{ message.series_slug }}/{{ message.slug }}" class="block">
+          <a href="/series/{{ series_slug }}/{{ message.slug }}" class="block">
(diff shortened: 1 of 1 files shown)


live.html
```diff
@@ -267,14 +267,18 @@ <h3 class="collection-header push-bottom">
   <div class="row">
     <div class="card-deck carousel" data-crds-carousel="mobile-scroll">
       <div class="feature-cards card-deck--expanded-layout">
-        {% assign recent_messages = site.messages | sort: 'date' | reverse | slice: 0,4 %} {% for message in recent_messages %}
+        {% assign recent_messages = site.messages | sort: 'date' | reverse | slice: 0,4 %} 
+        {% for message in recent_messages %}
+        {% for series in site.series %}
+          {% assign series_slug = series.slug | where: "videos.slug", message.slug %}
```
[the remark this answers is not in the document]
agause: @jasoncallibrity Can we give a message the ability to know the series to which it belongs?

live.html
```diff
@@ -267,14 +267,18 @@ <h3 class="collection-header push-bottom">
   <div class="row">
     <div class="card-deck carousel" data-crds-carousel="mobile-scroll">
       <div class="feature-cards card-deck--expanded-layout">
-        {% assign recent_messages = site.messages | sort: 'date' | reverse | slice: 0,4 %} {% for message in recent_messages %}
+        {% assign recent_messages = site.messages | sort: 'date' | reverse | slice: 0,4 %} 
+        {% for message in recent_messages %}
+        {% for series in site.series %}
```
lukemoderwell: Not sure that you need this extra loop. You could do the `assign` for a series and then set line `276` to `series.slug`.

lukemoderwell: Looks like this extra loop may be messing up some stuff. Deploy preview looks like it has the same image for all 4 messages.

jasoncallibrity: Nah, they literally had the same image in contentful. I mixed it up

jasoncallibrity: I'm using the slug from the message to find the series it belongs to. I have to loop through the series and compare the message slug to the videos slug until I get a match. Unless I'm missing something?

live.html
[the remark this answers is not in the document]
jasoncallibrity: I can add  {% break %} to it as I discovered it while we were talking to this existing code.

Do your second point, I will do some digging tomorrow to see if this is best practice. To be honest I'm not 100% sure. Thanks for calling it out!

PR #114, opened 2018-09-19
US14969: Critical Path CSS PoC

lukemoderwell:
PoC: Decided to manually inline homepage styles.

### DIFF
--- a/_includes/_head.html
+++ b/_includes/_head.html
@@ -32,4 +32,12 @@
   <meta name="theme-color" content="#ffffff">
 
   {% asset application.scss !integrity %}
+  <style>
+    {% if page.title == "Home" %}
+      {% capture home_css %}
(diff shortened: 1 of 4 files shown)


_includes/_head.html
```diff
@@ -32,4 +32,12 @@
...
 
   {% asset application.scss !integrity %}
+  <style>
+    {% if page.title == "Home" %}
+      {% capture home_css %}
+        {% include _critical-css/_homepage.scss %}
+      {% endcapture %}
+      {{ home_css | scssify | strip_newlines %}}
+    {% endif %}
+  </style>
```
jasoncallibrity: Shouldn't the entire `<style></style>` tag also be wrapped inside the `{% if %}`

lukemoderwell: good catch

PR #125, opened 2018-09-24
DE5901: Live Page Connect Section

lukemoderwell:
tweak html to match old look


live.html
```diff
@@ -122,11 +122,12 @@ <h2 class="text-white font-size-h3">crossroads connect</h2>
         <div class="col-md-5 soft">
           <article>
             <div class="media">
-              <div class="media-img-wrapper">
-                <div class="img-size-6 img-border">
```
jasoncallibrity: Yeah I remember making it 6 so that it fit, I knew we'd need to take it back to 7

PR #163, opened 2018-10-08
US15255: Redirect Unauthenticated Users

lukemoderwell:
listen for custom event, redirect url set


_assets/javascripts/lib/set-redirect-url.js
```diff
@@ -1,3 +1,6 @@
-window.onload = function() {
-    document.cookie = "redirectUrl=" + encodeURIComponent(window.location.href) + ";domain=.crossroads.net;path=/";
-};
\ No newline at end of file
+window.onload = function () {
+  var redirectUrlSet = new Event('redirect-url-set');
```
jasoncallibrity: add `is` predicate to the variable name and we are good to go.

`isRedirectUrlSet`

PR #177, opened 2018-10-15
update for gemlock file

jasoncallibrity:
changed jekyll-placeholders to use git instead of github (was complaining about it being insecure)

changed jekyll-cloudsearch to use tag instead of master branch, also changed to "git:" instead of "github:"


jasoncallibrity:
Self-merge for configuration change to gemlock.

PR #180, opened 2018-10-17
Production deploy for CRDS-NET

jasoncallibrity:


_assets/javascripts/components/simple-fred.js
```diff
@@ -0,0 +1,5 @@
+if (document.querySelector('simple-fred')) {
+  var script = document.createElement('script');
+  script.src = '//api-int.crossroads.net/fred/js/simplefred.min.js';
```
jasoncallibrity: Nothing with forms from Falcons is going live, so this being hard coded to int isn't currently an issue. There is much more effort required to make forms work

tcmacdonald: You can use the `CRDS.media.prefix` global to construct the environmentally specific subdomain. For example... 

```suggestion
  var prefix = CRDS.media.prefix == '' ? '' : '-' + CRDS.media.prefix;
  script.src = '//api'+ prefix +'.crossroads.net/fred/js/simplefred.min.js';
```

Not tested but should do the thing.

jasoncallibrity: We will definitely be testing this when we actually implement the simplefred stuff, but that's still a bit off.

PR #182, opened 2018-10-18
Locations happenings

adamtgeorge:


_assets/javascripts/views/locations.js
```diff
@@ -0,0 +1,12 @@
+new CRDS.CardCarousels('[data-crds-carousel]');
+new CRDS.CardFilters();
+
+function scrollToSection() {
+  var section = document.getElementById('happenings')
```
jasoncallibrity: Lets chat on this. If you take a look in the `live.js` file you'll see the following:

```  
  //Smooth scroller
  $(document).ready(function(event) {
    $('[data-smooth-scroll-to]').click(function(event) {
      event.preventDefault();    
      var targetId = $(this).data('smooth-scroll-to');
      var target = document.getElementById(targetId)
      if (target) {
        var scrollTo = $(target).offset().top;
        var offset = parseInt($(this).data('smooth-scroll-offset'));
        if (!isNaN(offset) && offset > 0) { scrollTo -= offset; }
        $('html, body').animate({ scrollTop: scrollTo }, scrollTo / 2);
      }
      return true;
    });
  });
```

If we are using this in more than one place now I'll give you a hand abstracting it out so we can make use of it in multiple places.

_layouts/location.html
```diff
@@ -124,11 +125,11 @@ <h4 class="card-title font-family-condensed-extra font-size-h3 text-uppercase">{
 
 <!-- Happenings -->
 {% if location_happenings.size > 0 %}
-<div class="container">
+<div class="container" id="happenings">
```
jasoncallibrity: we will add a `data-smooth-scroll-to="happenings"` to facilitate using the smooth scroller

PR #183, opened 2018-10-18
shenney/CYP - Series validation

s-henney:
- Adds a Cypress test suite to verify the current series displayed on the homepage, locations pages and media/series are all match the current series in Contentful

TODO

- Add test to check current series on /live (currently breaks Cypress)
- Add data-automation-ids to replace long css element finders
- Move the media/series test to crds-media to keep it connected to it's code base


cypress/integration/homepage/home_page_series_spec.js
```diff
@@ -0,0 +1,56 @@
...
+    it('Checks main current series: title, dates, image and description', function () {
+        cy.get('.current-series').as('currentSeriesBlock').should('be.visible');
+
+        const startDate = moment(content.currentSeries.starts_at);
+        const endDate = moment(content.currentSeries.ends_at);
+
+        cy.get('@currentSeriesBlock').find('div.container > div > div').then(($textBlock)=> {
+            expect($textBlock.find('h3')).to.have.text(content.currentSeries.title);
+            expect($textBlock.find('.card-subtitle')).to.have.text(`${startDate.format('MM.DD.YYYY')} — ${endDate.format('MM.DD.YYYY')}`);
+            expect($textBlock.find('.card-text > div > p')).to.have.text(content.currentSeries.description);
```
candrews-thrive: I am just curious why you did not use cypress's builtin assertions?

s-henney: I like the structure. All the child element tests are grouped together inside the parent element, so it's easy to tell what should be related on the page. It also seems like Cypress encourages creating "then" blocks like this.
Other than that, I don't have a compelling reason other than consistency with my own code.

PR #193, opened 2018-10-24
DE5945 footer spacing issue

jasoncallibrity:
added </br> to all layouts that have a footer to create proper padding. Will require each new layout we add to have this </br> tag added.

### DIFF
--- a/_layouts/container-fluid.html
+++ b/_layouts/container-fluid.html
@@ -3,5 +3,4 @@ layout: default
 ---
 {% include header.html %}
 {{ page.body | markdownify }}
-</br>
-{% include footer.html %}
+<div class="push-top">{% include footer.html %}</div>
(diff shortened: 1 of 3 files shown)


_layouts/container-fluid.html
```diff
@@ -3,4 +3,5 @@
 ---
 {% include header.html %}
 {{ page.body | markdownify }}
+</br>
 {% include footer.html %}
```
lukemoderwell: What about wrapping this in some `div` with a utility class? Not really that semantic either but feels slightly more in line with what we are trying to do.

```suggestion
<div class="push-top">{% include footer.html %}</div>
```

lukemoderwell: more context: https://stackoverflow.com/questions/1726073/is-it-sometimes-bad-to-use-br

PR #195, opened 2018-10-24
shenney/CYP- Automation ids for series

s-henney:
- added some data-automation-id tags to make it easier to target elements
- created the real /live test


cypress.json
```diff
@@ -1,4 +1,5 @@
 {
   "projectId": "dnbm5t",
-  "baseUrl": "https://int.crossroads.net"
+  "baseUrl": "http://127.0.0.1:4000/",
+  "numTestsKeptInMemory": 0
```
candrews-thrive: Does the above line "fix" memory leak with live?

s-henney: Nope. The only workaround I've found is to use cypress run instead of cypress open

PR #237, opened 2018-11-16
shenney/CYP - Cypress-Github integration (take 2)

s-henney:
- Added some files so that our Cypress tests will run when a PR is created, but should not prevent merge on failure. We're using Travis-ci to host/run the Cypress container and the tests'll run against the Netlify preview build.
- Only files changed or added to are cypress/kickOffCypress.sh, netlify.toml, travis.yml. The rest are line ending changes.

## TODO
- Still need to set some environment variables in Travis and Netlify


s-henney:
The "continuous-integration/travis-ci/push" check is our Cypress tests running against the Netlify preview. 
The two "Travis CI - *" checks can be ignored - they were misconfigurations I think I've fixed.

s-henney:
Most recent Travis tests failed due to image in Contentful with spaces in the name. Issue is being fixed in another PR.

PR #242, opened 2018-11-19
Extended component-header class for legacy support

bzmillerboy:


bzmillerboy:
Original PR lived in crds-styles but moved to this repo. https://github.com/crdschurch/crds-styles/pull/346

PR #287, opened 2018-11-28
shenney/DE_CYP - Split undefined fix

s-henney:
- Fixed a bug to avoid calling .split on a potentially undefined value

### DIFF
--- a/cypress/support/Contentful/Models/SeriesModel.js
+++ b/cypress/support/Contentful/Models/SeriesModel.js
@@ -37,11 +37,15 @@ export class SeriesModel {
     }
 
     get startDate(){
-        return this.starts_at.split("T")[0]; //Remove time zone
+        if(this.starts_at !== undefined)
+            return this.starts_at.split("T")[0]; //Remove time zone
+        return this.starts_at;
     }
 
     get endDate(){
-        return this.ends_at.split("T")[0]; //Remove time zone
+        if(this.ends_at !== undefined)
+            return this.ends_at.split("T")[0]; //Remove time zone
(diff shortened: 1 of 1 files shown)


cypress/support/Contentful/Models/SeriesModel.js
```diff
@@ -37,11 +37,15 @@ export class SeriesModel {
...
     get startDate(){
-        return this.starts_at.split("T")[0]; //Remove time zone
+        if(this.starts_at !== undefined)
+            return this.starts_at.split("T")[0]; //Remove time zone
+        return this.starts_at;
     }
 
     get endDate(){
-        return this.ends_at.split("T")[0]; //Remove time zone
+        if(this.ends_at !== undefined)
```
lukemoderwell: just a style thing: you may want to put a space after `if` and before the first `(`

cypress/support/Contentful/Models/SeriesModel.js
```diff
@@ -37,11 +37,15 @@ export class SeriesModel {
     }
 
     get startDate(){
-        return this.starts_at.split("T")[0]; //Remove time zone
+        if(this.starts_at !== undefined)
```
lukemoderwell: could you use a ternary here?
```suggestion
return this.starts_at !== undefined ? this.starts_at.split("T")[0] : this.starts_at
```

cypress/support/Contentful/Models/SeriesModel.js
s-henney: 👍

PR #292, opened 2018-11-28
US15789: Refactor crds-net code to use "{{page.body|markdownify}} as an include

janssejl:
Added new includes file and am using it in 4 _layouts.


lukemoderwell:
@janssejl I would consider changing the title of the include to something like `_page-content`.

PR #295, opened 2018-11-29
[hotfix] Release - Fix Travis config for Cypress

s-henney:


s-henney:
Travis check failures are due to bugs in the tests (which will be fixed) or content differences, not the config changes in this PR.

PR #343, opened 2018-12-04
hotfix - christmas changes 12/05

adamtgeorge:
• add if else statement to show different info for kids club on
georgetown


adamtgeorge:
@lukemoderwell  - I'm not sure how to test if this is working. Georgetown isn't a location in contentint. And in the preview link christmas doesn't show up on georgetown at all (or uptown).

PR #348, opened 2018-12-05
fix int home page

adamtgeorge:
• update config file to point to correct pathway


adamtgeorge:
@lukemoderwell @seancdavis @zarahzachz - Perspectives is working properly here. Any help you guys could offer would be wonderful.

PR #359, opened 2018-12-06
US15788 - Implement new asset-pipeline

seancdavis:


_assets/stylesheets/application.scss
```diff
@@ -1,7 +1,25 @@
-@import "crds-styles-sprockets";
-@import "./fonts";
-@import "./utilities";
-// @import "./hotfix";
-@import "./components/modules";
-@import "./pages/modules";
-@import "./vendor/flickity";
+@import '~crds-styles/assets/stylesheets/variables';
```
lukemoderwell: I'm unfamiliar with this style of import. What's happening here?

purgecss.config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "content": ["_site/**/*.html"],
+  "css": ["_site/assets/application*.css"],
+  "whitelist": []
```
lukemoderwell: We ran into some issues with purgecss in the past where we had to chuck a bunch of elements (dropdowns, btns, etc that are created via JS) onto this whitelist. Not the biggest deal but was fairly tedious when Olga would find regressions on legacy pages...

lukemoderwell: I guess tl;dr; do you have any tips for keeping a good whitelist?

_assets/stylesheets/application.scss
lukemoderwell: First guess, this is a another Gem/package in the ☁️ ?

seancdavis: @lukemoderwell The tilde represents a path to `node_modules`, using [this library](https://www.npmjs.com/package/node-sass-tilde-importer). That way we don't have to bring in vendor files but can use their npm package and still be able to use the sass `@import` idiom.

purgecss.config.json
seancdavis: @lukemoderwell Unfortunately, not a super easy way to do this. It's a tedious and manual process to add whitelisted classes to the config.

The other option you have is to build out a dummy page that has all the markup that isn't getting picked up. If you hide that from the nav but include it in the build, then theoretically those styles should get picked up by purgecss. And if you wanted to take it a step further you could write a hook to delete the file from the build after the build is complete.

That being said, I've found just adding the classes you need to the whitelist is the easiest way to go, as tedious as it may be.

seancdavis:
@lukemoderwell As far as performance is concerned, I saw great benefits on crds-media. We were seeing build times from 20-35 seconds. Since implementing this, build times have been closer to 10-12 seconds when building assets and 6-9 seconds when not needing to build assets.

Unfortunately, I haven't seen the same level of benefits in this project. My best guess is that is because:

1. There are so many ad-hoc scripts you're requiring that be built, so the actual asset build process is slow because it's compiling with Babel several times.
2. There's something else going on slowing down build times. Without assets, the builds in crds-net are 3-4x slower than crds-media. So there's probably something else going on.

Nevertheless, the benefits are still great. I'll take 20-30 seconds over the 78 I saw yesterday when testing something on the home page.

PR #369, opened 2018-12-07
feature/REF - Make Windows Happy Again

s-henney:
TL;DR Made some name changes so this repo can be served from Windows
- Contentful promo filesnames changed from :%S to .%S
- Colons and question marks removed from _page names

_shared_global_header.html was updated automatically from "bundle exec jekyll crds"


s-henney:
@lukemoderwell This was probably caused by bundle exec jekyll crds - I didn't manually change anything. The Shared Header is pulled from a content block though, so these changes shouldn't really affect anything.

PR #377, opened 2018-12-10
Add group leader proxy netlify

janssejl:


netlify.toml
```diff
@@ -13,15 +13,19 @@
 [context.branch-deploy.environment]
   CRDS_PEOPLE_DOMAIN = "crdspeople.netlify.com"
   CRDS_MAP_DOMAIN = "crdsmap.netlify.com"
+  CRDS_GROUP_LEADER_DOMAIN = "dev-group-leader.netlify.com"
```
lukemoderwell: I think you could just add this before your [build] command which runs on all builds regardless of context. I guess I should also make that same comment to map and people as well.

PR #378, opened 2018-12-11
Home Page - add data-optimize-bg-img

adamtgeorge:


adamtgeorge:
@lukemoderwell @bzmillerboy - Hoping this fixes the slow performance on INT.

PR #404, opened 2018-12-14
DE6218: Explore/Atrium Events JS Issues

lukemoderwell:
**Problem**
Atrium Events and Explore page JS was not added into the new JS config

**Solution**
Added scripts


lukemoderwell:
@seancdavis i'm seeing some weird behavior on the explore page with some minified JS (animations.min). Does the asset pipeline transform already minified assets?

seancdavis:
> @seancdavis i'm seeing some weird behavior on the explore page with some minified JS (animations.min). Does the asset pipeline transform already minified assets?

@lukemoderwell If the files are in the `files` array then they are compiled with Babel. If you add them instead to the `deps` array, they won't be transpiled.

PR #414, opened 2018-12-17
US15986 System Pages

tcmacdonald:
This PR adds a static copy of system-pages at the following URL and updates Cloudsearch with the contents of that file on each build... 

https://5c191175e4c8be0007c4b73a--int-crds-net.netlify.com/system-pages.json


_plugins/hooks/cloudsearch.rb
```diff
@@ -0,0 +1,44 @@
+require 'json'
+enabled = ARGV.include?('--update-search-index')
```
lukemoderwell: not familiar with `ARGV` but read up on it [here](https://stackoverflow.com/questions/13329132/whats-the-point-of-argv-in-ruby). Cool stuff

PR #429, opened 2018-12-19
hotfix - home page stream

adamtgeorge:


adamtgeorge:
@seancdavis @lukemoderwell - just need another set of eyes

PR #543, opened 2019-01-09
hotfix-superbowl-2019

adamtgeorge:
• list KC, FI serve links
• FB event links
• service times
• hide may trip 2


lukemoderwell:
@adamtgeorge closing this and reopening so that we can get a fresh netlify build

PR #552, opened 2019-01-10
Hotfix/rel more

adamtgeorge:


lukemoderwell:
@janssejl @j-brammer this build seems to failing due to release not having Sean's changes. Any thoughts on what we do?

lukemoderwell:
@adamtgeorge nothing that you did. Just has to do with some meta issues we've been seeing.

j-brammer:
The reason this is failing is due to meta references to objects in int. Since there is no real risk with this going to release to see what happens. The changes for meta objects have not been rolled up to demo or prod yet so there shouldn't be any issues.

PR #605, opened 2019-01-22
US15897 - Remove Maestro and add 404

janssejl:
## Problem
Maestro bad. 404 rom netlify not being served.

## Solution
Removing maestro redirect and changing 404 to be served by netlify


404.html
```diff
@@ -35,4 +181,56 @@ <h2 class="tagline push-bottom">We're sorry. This page doesn't seem to exist.</h
...
+      },
+      streamspotId: 'crossr30e3',
+      streamspotKey: 'a0cb38cb-8146-47c2-b11f-6d93f4647389',
+      streamspotPlayerId: '',
+      env: {
+        "gatewayServerEndpoint": 'https://gatewayint.crossroads.net/gateway/'
+      }
+    };
+  </script>
+  <script type="text/javascript" src="/assets/application-14d4a79cc6a1ae82db11930c.js" ></script>
```
lukemoderwell: Don't think we want these preprocessed assets here. We should be able to use the frontmatter as long as we throw that permalink up there.

PR #620, opened 2019-01-24
[CYP] US13857 - Pages not served by Maestro tests

s-henney:
## New test coverage & improvements
*Added some tests to make sure some pages do not revert to be served from Maestro.*


cypress/integration/infrastructure/not_served_by_Maestro_spec.js
```diff
@@ -0,0 +1,103 @@
+function responseShouldNotBeFromMaestro(response){
+  expect(response.headers['x-varnish']).not.to.exist;
+
+  if(response.headers.via != undefined){
+    expect(response.headers.via).not.to.contain('varnish');
+  }
+}
+
+function pageShouldNotBeFromMaestro(response){
+  expect(response.body).not.to.contain('meta name="Maestro"');
```
j-brammer: We should probably test the positive versus the negative here

instead of seeing if it doesn't have Maestro for crds-net pages lets test to see if it has Netlify

This below code exists in _head.html in the crds-net project
`<meta name="Netlify" content="This page was rendered with Netlify">`

s-henney: Netlify check added.

PR #642, opened 2019-01-31
[CYP] Authenticated test fix

s-henney:
## Problem
*A set of tests weren't using the defined test user.*
*There's a big modal in /camps that's blocking the profile icon.*

## Solution
*Send the password variable to Travis.*
*Skip the /camp test until we can work around the modal better.*


cypress/integration/sharedHeader/crossroads_logo_redirect_authspec.js
```diff
@@ -1,4 +1,5 @@
 import { RouteValidator } from '../../support/RouteValidator';
+import { fred_flintstone } from '../../fixtures/test_users';
```
lukemoderwell: I need more imports like this in my life.

PR #653, opened 2019-02-04
USXXXX: Remove Search From 404

lukemoderwell:
## Problem
Our new search isn't able to be search via query params (for now).

## Solution
Remove search input on 404 in favor of a button that links to search

### Corresponding Branch
N/A

## Testing
Is there a button there? Does it link me to search?

### DIFF
--- a/404.html
+++ b/404.html
@@ -9,21 +9,7 @@ permalink: /404.html
     <h2 class="tagline push-bottom">We're sorry. This page doesn't seem to exist.</h2>
   </div>
   <div class="col-sm-6">
-    <p>Please use the navigation above or search here.</p>
-    <form role="form" action="{{ site.search_domain }}/search">
-      <div class="form-group">
-        <div class="input-group input-group-lg">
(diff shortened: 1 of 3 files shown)


lukemoderwell:
Waiting on search release...

404.html
```diff
@@ -9,21 +9,7 @@ <h1 class="text-lowercase page-header">Oops</h1>
...
-            <button type="submit" class="btn btn-blue">
-              <svg class="icon icon-1" viewbox="0 0 256 256">
-                <use xlink:href="/assets/svgs/icons.svg#search"></use>
-              </svg>
-            </button>
-          </span>
-        </div>
-      </div>
-    </form>
+    <a class="full-width btn btn-primary" href="/search">Search Crossroads.net</a>
```
j-brammer: Did we not find a way to pass a param into search? Seems like a nice thing to have on the 404 page...I'm not going to stop this from going in due to that, but it seems like passing a param from here into search would be fairly easy with angular?

lukemoderwell: @j-brammer have this story on the backlog to handle: https://rally1.rallydev.com/#/41662702253d/detail/userstory/283954837584

PR #667, opened 2019-02-06
[CYP] - Fixed 404 tests after change

s-henney:
## Problem
*404 search field was removed.*

## Solution
*404 validation tests now target the search button.*


404.html
```diff
@@ -9,17 +9,19 @@ <h1 class="text-lowercase page-header">Oops</h1>
     <h2 class="tagline push-bottom">We're sorry. This page doesn't seem to exist.</h2>
   </div>
   <div class="col-sm-6">
-    <a class="full-width btn btn-primary" href="/search">Search Crossroads.net</a>
+    <a class="full-width btn btn-primary" href="/search" data-automation-id="404-search-button">Search Crossroads.net</a>
```
lukemoderwell: In the future, I'll try to do a better job doing this in the dev.

lukemoderwell: @s-henney is there a convention that you usually use for data-automation ids?

PR #700, opened 2019-02-25
US16586 Health Check

tcmacdonald:
## Problem
Intermittent build failures sometimes result in a zero-length index.html file. 

## Solution
This PR adds a new bash script to the end of our build command that looks for the presence of a string in `_site/index.html` and if it is not present, throws a non-zero exit code which will cause the build to fail. 

## Testing
The first build on this PR was designed to fail. You can review [the log output](https://app.netlify.com/sites/int-crds-net/deploys/5c74365a04672f00086c6d6f) but it basically reports this... 

```1:40:40 PM: <\!-- we are crossroads. --> not found in ./_site/index.html```

This messaged was logged (and the build failed) due to the trailing period which does not exist in `_site/index.html`. The subsequent build corrects the string we're looking for, thus the build should pass.

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,39 +1,33 @@
 <!DOCTYPE html>
 <html lang="en">
-{% include _head.html %}
+  {% include _head.html %}
 
-<body class="crds-styles">
-  {% include _gtm-body.html %}
-  {% if page.requires_auth %}
-  {% include _preloader.html %}
-  {% endif %}
-  {{ content }}
-  {% include _monetate-body.html %}
-  <script>
-    var CRDS = {
-      media: {
-        cms: '{{ site.shared_header.cms }}',
-        app: '{{ site.shared_header.app }}',
-        img: '{{ site.shared_header.img }}',
-        prefix: '{{ site.shared_header.prefix }}'
-      },
-      streamspotId: '{{ site.streamspotId }}',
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: _layouts/default.html [hunk 1]
```diff
@@ -1,39 +1,33 @@
...
+        streamspotPlayerId: "{{ site.streamspotPlayerId }}",
+        env: {
+          gatewayServerEndpoint: "{{ site.gateway_server_endpoint }}"
+        }
+      };
+    </script>
+    {% javascript_link_tag application %} {% javascript_link_tag
+    application_deferred async %} {% if page.masonry_js %} {%
+    javascript_link_tag masonry_deferred async %} {% endif %} {% if page.title
+    == "Atrium Events" %} {% javascript_link_tag events async %} {% endif %}
```
j-brammer: Y U h8 formatting?

_layouts/default.html
```diff
@@ -1,39 +1,33 @@
...
 
+    <!-- we are crossroads -->
```
j-brammer: could we put this on the actual index.html?

Just for clarity, we decided as a team that / will NOT be coming to contentful. So instead of every page having that, only the one that decides to be blank will have it.

_layouts/default.html
tcmacdonald: Oh sorry, Prettier did this. I will revert.

PR #737, opened 2019-03-06
[CYP] Better /live coverage and some bug fixes

s-henney:
## New test
- Added coverage for /live "Watch This Weeks Service" button URL and new page load

## Fixes
- Fixed /live Past Weekend message tests to verify absolute URL
- Changed /serve test to /volunteer to match Prod and Demo, but skipping it until Contentful is refreshed in Int again. This test coverage is somewhat redundant, so skipping for a few weeks is low risk.

## Improvements
- Absolute and relative URLs stored in SeriesModel. Absolute URLs stored in MessageModel.
- Removed unnecessary tests confirming page wasn't served from Monetate
- failOnStatusCode now optional for ContentfulApi calls.


cypress/Contentful/ContentfulApi.js
```diff
@@ -22,12 +22,12 @@ class ResponseWrapper {
 }
 
 export class ContentfulApi {
-  static getEntryCollection(query) {
+  static getEntryCollection(query, failOnStatusCode=true) {
```
j-brammer: It's not super clear what the purpose of `failOnStatusCode` is. Possible to add something to make it more clear or a comment? I see it's reused a lot so that might be a challenge

s-henney: `failOnStatusCode` is the name Cypress uses - I was just copying them. Luke asked about this too though, so I'll think of a better name and add some comments.

PR #775, opened 2019-03-19
Demo deploy 2005

j-brammer:
## Problem
*Document current behavior and why it's undesirable here.*

## Solution
*Document how the behavior was fixed or enhanced here.*

### Corresponding Branch

## Testing


spec/fixtures/redirects.csv
```diff
@@ -1,3 +1,3 @@
 http://crossroads.net/*,https://www.crossroads.net/:splat,301!,master
-http://${env:CRDS_APP_DOMAIN}/*,https://${env:CRDS_APP_DOMAIN}/:splat,301!
+http://int.crossroads.net/*,https://int.crossroads.net/:splat,301!
```
s-henney: Does this need to be changed for prod?

j-brammer: No. this test should only run against one environment due to how it's made. It can't be made dynamic as far as I can tell

j-brammer: In fact, it being dynamic is what broke it

PR #780, opened 2019-03-20
De6564 feature media entries

phritolay:
## Problem
There are 5 entries in (net) Landing Page features in Contentful, only 4 are displayed.

## Solution
update features slice and use `.docs` instead of `.entries`


index.html
```diff
@@ -142,7 +142,7 @@ <h2 class="flush-top font-size-base text-gray-light">
       </h2>
 
       {% if managed_features.size > 0 %}
-        {% assign features = managed_features | slice: 0, 4 %}
+        {% assign features = managed_features | slice: 0, 100 %}
```
j-brammer: would this work?
`{% assign features = feature_obj.docs | slice: 0, 100 %}`

Then you could remove line 133

PR #803, opened 2019-03-27
DE6603 - Featured Media Links

seancdavis:
Only call `get_doc` for featured items if url property is not already there.

These guys are not pointing to the correct URLs, but there are redirects in place that make it work. However, it is using the paths that come from Jekyll, so I feel like it is a reasonable approach (for now).


seancdavis:
@JenEllis44 The super bowl link is not working for an unrelated issue. Try clicking the second link in the list.

PR #818, opened 2019-04-03
[CYP] Dedupe tests with Cypress's "test generator" functionality

s-henney:
- Cypress can auto-generate tests given an array of parameters. Converted similar tests to use this functionality.
- Fixed some bugs noticed in demo tests
- Split location_search_spec into two files


cypress/integration/live/live_latest_message_spec.js
```diff
@@ -38,24 +18,27 @@ describe('Testing the Past Weekends section on the Live page:', function () {
     });
   });
 
-  it('Most recent message card should contain title, image, description and link', function () {
-    const index = 0;
-    check_message_card_content(index, recentMessages[index]);
-  });
+  [0,1,2,3].forEach((index) => {
```
lukemoderwell: What's up with this array? Picking the first 4?

lukemoderwell: Never mind. I got it 👌

PR #842, opened 2019-04-15
US16851 - Home page improvements

tallented:
## Problem
* Require small changes to home page

## Solution
* Change “recent articles” -> “recent”
* “latest articles” -> “latest” when no recent articles defined
* Close gap between latest podcast episodes and explore our topics

### Corresponding Branch
https://rally1.rallydev.com/#/41662702253d/detail/userstory/295417595880

## Testing
View homepage at various widths

### DIFF
--- a/index.html
+++ b/index.html
@@ -167,7 +167,7 @@ paginate:
         {% if managed_features.size > 0 %}
         featured
         {% else %}
-        latest articles
+          latest
         {% endif %}
       </h2>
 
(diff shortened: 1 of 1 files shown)


tallented:
@JenEllis44 There is a note in the description of the story that two different titles should be conditionally displayed

```
@Chris: update to #1 --
*Featured* - label when the featured content model is in use
*Latest* should be when featured content model is NOT in use
```

Currently the site is showing "Featured" because in Contentful featured content has been.  If there wasn't any defined featured content then it would display "Latest" as Adam's note requests.  I'm not actually sure how to test for this?

index.html
```diff
@@ -167,7 +167,7 @@ <h2 class="flush-top font-size-base text-gray-light">
         {% if managed_features.size > 0 %}
         featured
         {% else %}
-        latest articles
+          latest
```
tallented: This is the change to display "latest" but it is only triggered when there are no managed features

JenEllis44: Makes sense! Thanks for letting me know.

PR #863, opened 2019-04-25
Feature/us16854 netlify logs

j-brammer:
Changes:
Added two new environment variables

- LOGZIO_API_KEY
- APPLICATION_NAME

When a build fails, only the error messages will be sent to Logzio. You can search for these by typing the following into the search bar
`type:"APPLICATION_NAME"` where application name is the variable set in Netlify or locally.

`1> /dev/null` signifies that STDOUT will be redirected to /dev/null and will not be seen
`2> buildlog.txt` signifies that STDERR will be redirected to our build log file to be shipped

Logzio.sh only runs if there is an error. So no logs will be shipped on a successful build.

The buildlog.txt file will be removed so it does not display on the site


redirects.csv
```diff
@@ -1,5 +1,3719 @@
 http://crossroads.net/*,https://www.crossroads.net/:splat,301!,master
 http://${env:CRDS_APP_DOMAIN}/*,https://${env:CRDS_APP_DOMAIN}/:splat,301!
+/real,https://int.crossroads.net/media/series/real-encounters-with-god,302
```
tcmacdonald: 3700 new redirects? This is a mistake, right?

bin/build-command.sh
```diff
@@ -0,0 +1,14 @@
...
+# Ships logs to logz.io if there is a failure
+
+(bundle exec rspec 1> /dev/null 2>> buildlog.txt &&
+./bin/contentful-redirects 1> /dev/null 2>> buildlog.txt &&
+./bin/netlify-redirector 1> /dev/null 2>> buildlog.txt &&
+bundle exec jekyll crds 1> /dev/null 2>> buildlog.txt &&
+bundle exec jekyll contentful -f 1> /dev/null 2>> buildlog.txt &&
+bundle exec jekyll build -- --update-search-index 1> /dev/null 2>> buildlog.txt &&
+bash ./cypress/kickOffCypress.sh 1> /dev/null 2>> buildlog.txt &&
+./bin/health-check.sh "we are crossroads" 1> /dev/null 2>> buildlog.txt) ||
```
tcmacdonald: Can you group the output of these commands to make your script a little more readable? 

```
{
  bundle exec rspec &&
  ./bin/contentful-redirects &&
  ...
} 1> /dev/null 2>> buildlog.txt
```

j-brammer: And I learn another Bash thing today. Thanks Taylor. Will change

redirects.csv
j-brammer: 100% a mistake

tcmacdonald:
Yea, but if its not an error thats useful information. For example, we used to log the number of records pulled from Contentful, the parsed output of redirects.csv, etc.

redirects.csv
j-brammer: @tcmacdonald Fixed

bin/build-command.sh
j-brammer: @tcmacdonald Fixed

j-brammer:
@tcmacdonald Concern addressed. It now outputs 100% of the log to Netlify and uses Logzio as a backup when it truncates logs. the Logzio side is pretty ugly because of the lack of configurability using curl to send logs but it's better than nothing. If I could install things on the build image to configure it to ship to our logzio setup I would. The one way I could outputs the key in plaintext in the build log and requires sudo permissions to configure.

PR #872, opened 2019-05-06
[CYP] Reporting to Cypress dashboard

s-henney:
- This feature was turned off in April when we ran out of dashboard storage, which auto-failed tests. Turning it back on here.
- Please ignore the failing Travis checks, this is expected and was just used for testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,8 @@ language: node_js
 node_js: 10
 cache: npm
 install: npm ci
-jobs:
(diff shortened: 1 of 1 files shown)


s-henney:
@lukemoderwell Unfortunately yes. We're looking to get on a plan with a higher limit soon though, so hopefully we won't have that issue.
We have some other workarounds options, but I'm just going to kick that can down the road a little longer for now.

PR #876, opened 2019-05-13
give /app namespace class

jakecatron:
## Problem
The stylesheet for /app currently uses global style overrides which will impact other pages in crds-net.

## Solution
Create a namespace class in order to scope these style overrides to only /app.

### DIFF
--- a/_assets/stylesheets/pages/_app.scss
+++ b/_assets/stylesheets/pages/_app.scss
@@ -1,4 +1,4 @@
-
+.app-page-styles {
 
 /* iPhone X & Pixel 2 XL */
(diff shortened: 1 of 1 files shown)


_assets/stylesheets/pages/_app.scss
```diff
@@ -1,4 +1,4 @@
-
+.app-page-styles {
 
 /* iPhone X & Pixel 2 XL */
 @media (max-height: 896px) and (max-width: 414px) {
```
lukemoderwell: just a style note: you can indent all the styles inside the parent class.

PR #882, opened 2019-05-23
Replace Stream Player

lukemoderwell:


.env-sample
```diff
@@ -7,8 +7,7 @@ CRDS_CMS_SERVER_ENDPOINT=
 PUSHPAY_GIVE_ENDPOINT=
 IMGIX_SRC=
 IMGIX_DOMAIN=
-STREAMSPOT_API_KEY
-STREAMSPOT_ID=
+STREAM_SCHEDULE_ENDPOINT=https://8k97vbzbrk.execute-api.us-east-1.amazonaws.com/int/streamSchedule
```
j-brammer: do we want the actual value stored in github?

_assets/javascripts/components/bitmovin.js
```diff
@@ -0,0 +1,248 @@
...
+class BitmovinManager {
+  constructor(bitmovinConfig) {
+    this.isCard = bitmovinConfig.isCard;
+    this.isStream = bitmovinConfig.isStream;
+    this.videoDuration = Number(bitmovinConfig.duration) * 1000 ;
+    this.timezoneStr = 'America/New_York';
+    this.container = document.getElementById(`${bitmovinConfig.id}`);
+    this.countdown = new CRDS.Countdown();
+    this.playerConfig = {
+      key: '224f523d-e1ba-4f96-ad4d-96365f461c93',
```
j-brammer: Can we use an environment variable here?

_assets/javascripts/components/bitmovin.js
```diff
@@ -0,0 +1,248 @@
...
+      playback: {
+        autoplay: this.getAutoPlay(),
+        muted: this.getIsMuted()
+      },
+      analytics: {
+        key: '01e90136-7623-4df4-b0d9-a3d975b00258',
```
j-brammer: same here?

j-brammer: After thinking about it, we might want to let bitmovin know we exposed these (PR's are forever) and see if that has any implications for driving up our bill if someone were to misuse it.

_assets/javascripts/components/bitmovin.js
```diff
@@ -0,0 +1,248 @@
...
+  getQualityLabels(data) {
+    let resolution = '';
+    let kbps = Math.round(data.bitrate / 1000);
+
+    // convert to megabits if applicable
+    let bitrate =
+      kbps > 1000 ? `${(kbps / 1000).toFixed(1)} mbps` : `${kbps} kbps`;
+
+    if (data.height <= 240) {
+      resolution = '240p';
```
j-brammer: should these be labeled `SD` as well? the standard YouTube uses is anything below 720p just doesn't get a definition label. they don't even use SD.

YouTube Options

index.html
```diff
@@ -53,7 +53,7 @@
           <div class="soft-quarter-top font-size-base flush-bottom" data-automation-id="live-message-description">
             Join the live stream hourly, on the hour, from 8am-10pm (EST).
           </div>
-          <a href="/live/stream/" class="btn btn-primary btn-sm push-half-top"
+          <a href="/live/stream?autoplay=true&sound=11" class="btn btn-primary btn-sm push-half-top"
```
j-brammer: I don't particularly care, but we could do the following now
`{{ "/live/stream" : media_url }}`

then if we ever change the parameter we can just change it in the filter and not have to find all these references

this would require you to merge development into your branch

stream.html
```diff
@@ -0,0 +1,153 @@
+---
+layout: no-header-no-footer
+permalink: video-test
+---
```
j-brammer: do we want to keep this file or let the file live in contentful?

.env-sample
lukemoderwell: maybe not. I'll remove

PR #910, opened 2019-06-05
remove specific spark talks

adamtgeorge:
• deleting old talks
• updates will be coming weekly when spark launches


adamtgeorge:
@lukemoderwell @j-brammer - I tried running autopick but it was including a other changes by Jake B. for some reason

PR #915, opened 2019-06-05
DE6830 sitemap exclusions

j-brammer:
## Problem
pages which are excluded from search are showing up in the sitemap.xml

## Solution
added hook to exclude pages from sitemap when search_excluded is true

### Corresponding Branch

## Testing

### DIFF
--- a/_plugins/hooks/sitemap.rb
+++ b/_plugins/hooks/sitemap.rb
@@ -1,0 +1,9 @@
+require 'pry'
+
+Jekyll::Hooks.register :documents, :pre_render do |doc|
+
+  if doc.collection.label == 'pages'
+    doc.data['sitemap'] = !doc.data['search_excluded']
+  end
+  
+end


_plugins/hooks/sitemap.rb
```diff
@@ -0,0 +1,9 @@
+require 'pry'
+
+Jekyll::Hooks.register :documents, :pre_render do |doc|
+
+  if doc.collection.label == 'pages'
+    doc.data['sitemap'] = !doc.data['search_excluded']
```
lukemoderwell: is `search_excluded` always going to true/false? My understanding is they are trinary (true, false, null). Is that going to bork anything?

lukemoderwell: @j-brammer I guess I'm really just wondering what happens when that radio isn't checked.

tcmacdonald: ☝️This a good catch @lukemoderwell. 

I would assume that anything not checked will return undefined, which would evaluate to true in this expression.

lukemoderwell: @tcmacdonald Jason just tested and you are correct. Undef will return `false` which, since we are taking the opposite, gives us the desired behavior. No radio selection = include in sitemap.

PR #916, opened 2019-06-06
US17046 - JSON Files

seancdavis:
Add shared header data files.

See crdschurch/crds-components#32.


seancdavis:
Decided to go a different direction with this. See crdschurch/crds-data for more info.

PR #924, opened 2019-06-12
Adding proxy rules for new okta signin

dillon-courts:


redirects.csv
```diff
@@ -58,4 +58,5 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /authors/*,/media/authors/:splat,302!
 /songs/*,/media/songs/:splat,302!
 /media/*,${env:CRDS_MEDIA_ENDPOINT}/:splat,200!
+/oktasignin,https://${env:CRDS_SIGNIN_ENDPOINT},200!
```
tcmacdonald: You forgot the`/:splat`.

dillon-courts: Do I need the splat? My understanding was that the splat was if you had additional paths appended to that route. /oktasignin is a deadend. Or is the splat for the query string params?

dillon-courts: I can change it to this - /oktasignin/*,https://${env:CRDS_SIGNIN_ENDPOINT}/:splat,200!

just wasn't sure it was necessary.

dillon-courts: For example - these 3 paths in the redirects file do not contain the splat:

/signin,https://${env:CRDS_ANGULAR_ENDPOINT}/signin,200!
/signout,https://${env:CRDS_ANGULAR_ENDPOINT}/signout,200!
/register,https://${env:CRDS_ANGULAR_ENDPOINT}/register,200!

j-brammer: @tcmacdonald @dillon-courts if there are url parameters going from one application to the other they will not be passed along.

tcmacdonald: @dillon-courts I just assumed there were other routes under `/oktasignin` that you might need to capture like '/oktasignin/forgot-password` etc. If not, no biggie. 

@j-brammer In the case of a proxy, query string params do indeed get passed through the  microclient. We're doing this on crds-map, [for example](https://int.crossroads.net/connect/?lat=39&lng=-84.51&zoom=9&filters=At%20Home%20Groups%3D&my=0). Let me know if I misunderstood.

dillon-courts: No additional routes that I'm aware of. And I tested the query string params with https://int.crossroads.net/register. They do get passed along. Thanks guys, I'm going to resolve this.

PR #963, opened 2019-06-24
Shared Header, Footer, and Snail Trail

seancdavis:


seancdavis:
This says it's failing, but it's a Netlify issue with the Runway project specifically, while the int builds are succeeding. And those are the ones we care about.

PR #968, opened 2019-06-25
DE6922 - Watch latest service button href not from media snippets anymore

s-henney:
## Problem
- "Watch the latest message" link on the homepage broke when Shared Header 2.0 was deployed

## Solution
- Changed the href source to  not use media snippets
- Updated automated tests

### DIFF
--- a/index.html
+++ b/index.html
@@ -314,7 +314,8 @@ paginate:
               <h4 class="component-header" style="line-height:.9em;"><span class="text-gray-lighter">Watch a
                   service</span><br><span class="text-white">online</span></h4>
               <span class="font-size-small">Pants optional. Coffee encouraged.</span><br>
-              <a class="btn btn-secondary btn-lg" href="{{ series.url }}"
+{% assign recent_series = home_series | first %}
+<a class="btn btn-secondary btn-lg" href="{{ recent_series.url }}"
                 data-automation-id="watch-series-button">Watch the latest service</a>
             </div>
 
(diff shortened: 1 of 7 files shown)


index.html
```diff
@@ -314,15 +314,15 @@ <h1 class="font-family-condensed-extra text-uppercase">What is Crossroads</h1>
               <h4 class="component-header" style="line-height:.9em;"><span class="text-gray-lighter">Watch a
                   service</span><br><span class="text-white">online</span></h4>
               <span class="font-size-small">Pants optional. Coffee encouraged.</span><br>
-              <a class="btn btn-secondary btn-lg" data-media-snippet="get_current_series_url:attr[href]"
+              <a class="btn btn-secondary btn-lg" href="{{ series.url }}"
```
lukemoderwell: @s-henney I don't think this is going to work. That series object only exists in certain `for` loop contexts and I don't think it's available here. We can probably create a global variable for you to access though.

lukemoderwell: maybe something like this?
```suggestion
   {% assign recent_series = site.series | sort: 'date' | reverse | first %}
   <a class="btn btn-secondary btn-lg" href="{{ recent_series.url }}"
```

s-henney: Would this work instead?
```suggestion
{% assign recent_series = home_series | first %}
<a class="btn btn-secondary btn-lg" href="{{ recent_series.url }}"
```
"home_series" is defined in the "Current Series" section and is what's used to display the current series card. I'd like to reuse that logic here if possible.

PR #1041, opened 2019-07-18
update styles for radio buttons

adamtgeorge:


adamtgeorge:
/worshiptraining/step2

PR #1072, opened 2019-08-02
AUTH FOR HOMEPAGE WOOH.

janssejl:


_assets/javascripts/lib/authentication.js
```diff
@@ -0,0 +1,39 @@
+
+let oktaConfig = {
+  clientId: CRDS.env.okta_client_id,
```
j-brammer: Is there a reason we are declaring clientId and issuer using {site.XXX} on auth-required and using CRDS.env here?

_assets/javascripts/lib/authentication.js
```diff
@@ -0,0 +1,39 @@
...
+    crdsAuth.CrdsAuthenticationProviders.Okta,
+    crdsAuth.CrdsAuthenticationProviders.Mp
+  ]
+}
+
+let authService = new crdsAuth.CrdsAuthenticationService(authConfig);
+
+authService.authenticated().subscribe(tokens => {
+  if (tokens)
+    document.getElementsByTagName("crds-site-happenings")[0].setAttribute("auth-token", tokens.access_token.access_token);
```
lukemoderwell: A consideration here: Maybe look for that prop in general and then update all components using this concept.

```
let authEls = document.querySelectorAll('[auth-token]')
authEls.forEach(...set this attr)
```

_assets/javascripts/lib/authentication.js
```diff
@@ -0,0 +1,39 @@
...
+  providerPreference: [
+    // Put these in the order you want the library to check for auth status
+    crdsAuth.CrdsAuthenticationProviders.Okta,
+    crdsAuth.CrdsAuthenticationProviders.Mp
+  ]
+}
+
+let authService = new crdsAuth.CrdsAuthenticationService(authConfig);
+
+authService.authenticated().subscribe(tokens => {
```
j-brammer: Seems like the only difference in authentication.js and auth-required.js is this little block of code here. 

Could we maybe key off the "auth required" contentful page attribute to decide which type of auth (full page required, or just component required) to use?

_assets/javascripts/lib/authentication.js
janssejl: done

_assets/javascripts/lib/authentication.js
janssejl: refactored.

_assets/javascripts/lib/authentication.js
janssejl: just goofed on not changing that file from {site.....} when I realized that didnt work. Refactored anyway.

PR #1134, opened 2019-09-05
de7120 update onsite groups snail trail

phritolay:
## Testing
https://deploy-preview-1134--int-crds-net.netlify.com/groups/onsite


onsite-groups.html
```diff
@@ -13,7 +14,7 @@ <h1>
     </svg>
   </h1>
   <p>
-    It’s true. We get by with a little help from our friends. Find yours here.
+    Itâ€™s true. We get by with a little help from our friends. Find yours here.
```
tcmacdonald: Nope.

phritolay: That's really weird, I didn't change that. 🤔 

Pushing up a commit to change it back.

PR #1150, opened 2019-09-17
Remove /giving redirect

adamtgeorge:
• Fix the navigation


adamtgeorge:
@j-brammer - The above didn't fix it but I deleted the redirect. Maybe I'm slow.

PR #1162, opened 2019-09-23
[CYP] US17424 - HP current message autoplays

s-henney:
- Tests that the current message card on the homepage autoplays (if Bitmovin)


cypress/integration/homepage/home_page_latest_message_spec.js
```diff
@@ -1,44 +1,70 @@
...
+    //Navigate
     cy.ignorePropertyUndefinedTypeError();
     cy.visit('/');
   });
 
-  it('Current Message title, description, and image should match Contentful', function () {
-    cy.get('[data-automation-id="message-title"]').as('title');
-    cy.get('@title').text().should('contain', currentMessage.title.text);
-    cy.get('@title').should('have.attr', 'href', relativeAutoplayURL);
+  it('Checks title, image, and "View latest now" button have correct link', () => {
```
j-brammer: question: does Cypress reuse the session in a way that might make it so you can break this test up?

have a
`Checks title has correct link`
`Checks image has the correct link`
`Checks "View Latest now" button has correct link`

Then you end up with smaller tests that might be more readable and more debuggable.

cypress/integration/homepage/home_page_latest_message_spec.js
```diff
@@ -1,44 +1,70 @@
...
+      cy.get('[data-automation-id="message-video"]').as('videoImagelink');
+      cy.get('@videoImagelink').should('have.attr', 'href', relativeAutoplayURL);
+
+      cy.get('[data-automation-id="watch-message-button"]').as('watchMessageButton');
+      cy.get('@watchMessageButton').should('be.visible');
+      cy.get('@watchMessageButton').should('have.attr', 'href', relativeAutoplayURL);
     });
+  });
 
+  it('Checks card image and, if Bitmovin video, player exists and video autoplays', () => {
```
j-brammer: Same deal here. I'm always in favor of smaller more granular tests if possible. If I remember correctly Cypress will use the same loaded page to check these things so you should be able to break these up

cypress/integration/homepage/home_page_latest_message_spec.js
s-henney: Unfortunately no, which is why you might notice me bundling more checks into a single test. Their before/after each processes run after each "it" block and seem to be pretty process heavy. A spec file similar to this one took 17s if tests were broken up into really small chunks but only 2s when more assertions were bundled together.
Cypress wants you to test multiple assertions in a single test, they even have an entry in their best practices section [here](https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion). The tl;dr is Cypress tests aren't unit tests, so shouldn't be structured in the same way.

j-brammer: gross. Understood

PR #1169, opened 2019-10-01
Feature/us17833 homepage logged in layout

bzmillerboy:


index.html
```diff
@@ -509,7 +509,6 @@ <h5 class="font-family-condensed-extra text-uppercase flush-ends"><a href="{{ po
     c.getEventStatus();
   }
 </script>
-{% javascript_link_tag autoplay async %}
-{% javascript_link_tag skeleton async %}
 
-<!-- we are crossroads -->
+{% javascript_link_tag autoplay async %}
+{% javascript_link_tag skeleton async %}
```
j-brammer:
```suggestion
{% javascript_link_tag skeleton async %}

<!-- we are crossroads -->
```

Add this back in. It's what we use to make sure the homepage isn't blank after build

see bin/health-check.sh

PR #1187, opened 2019-10-11
update link for Kids' Club to be crossroads.net page

adamtgeorge:


adamtgeorge:
@j-brammer - When will this be in PROD? Next week?

PR #1194, opened 2019-10-21
Feature/us18234 exclude igr categories

J-Byron:
### Problem 
- Categories that belong to IGR collection should be excluded from NET
- IGR collection object should also be removed


_config.yml
```diff
@@ -49,14 +49,14 @@ contentful:
     query: fields.category=common
     map:
       content_block: content
-  page:
-    query: "fields.collections[exists]=false"
```
tcmacdonald: Was this intentional?

_config.yml
```diff
@@ -49,14 +49,14 @@ contentful:
...
   podcast:
     query: "fields.exclude_from_crossroads[ne]=true"
   episode:
     query: "content_type=episode&fields.podcast.sys.contentType.sys.id=podcast&fields.podcast.fields.exclude_from_crossroads[ne]=true"
   video:
     query: "fields.exclude_from_crossroads[ne]=true"
+    map:
+      date: published_at
```
tcmacdonald: Was this intentional?

_config.yml
```diff
@@ -159,6 +159,10 @@ contentful:
     query: "fields.exclude_from_crossroads[ne]=true"
     map:
       date: published_at
+  category:
+    query: "fields.collections.sys.id[nin]=3F1707Eeiu50rMaxCf6wNN,3X7oFhglWTJGXy3TD5u5br,6XwgUaVRdsKzXrncprwJMt,3E8JlwCIG49c00bU7MXEuL"
+  collection: 
+    query: "fields.slug[ne]=isgodreal-com"
```
tcmacdonald: This seems a little brittle to me. Is there anyway to query for categories that have no collection assigned at all?

J-Byron: If I understand the question correctly, yes, this query string already INCLUDES categories with no collection assigned. This query includes all categories except those with collections ids of IGR.

_config.yml
J-Byron: I was having issues with builds/contentful because collections was removed from pages in development. This has since been reverted so I will in turn revert this.

_config.yml
seancdavis: For category, could you use what we've done for pages?

```
fields.collections[exists]=false
```

For collection, why bring them in at all? All we're using them for is the query. I think we could actually ignore them entirely and we should still be fine.

_config.yml
seancdavis: IIRC this would prevent videos without a date in the past from rendering on the front end, but I'm not sure why that's relevant to this work.

_config.yml
seancdavis: Not important, but now it just looks like you added some extra spaces.

PR #1220, opened 2019-11-19
updating css for buttons on locations page to make them aligned

angelicawink:
Updated CSS in order to align buttons on locations

**NOTE**: this also required updating HTML in one file (_locations-grid.html) because <crds-button/> on the "Anywhere" card was nested in the wrong div.


.env-sample
```diff
@@ -1,19 +1,15 @@
...
-CRDS_CMS_ENDPOINT=https://contentint.crossroads.net
-CRDS_GATEWAY_ENDPOINT=https://gatewayint.crossroads.net/gateway
-CRDS_MEDIA_ENDPOINT=https://mediaint.crossroads.net
-CRDS_SEARCH_ENDPOINT=https://int-crds-search.netlify.com
-AWS_ACCESS_KEY_ID=
-AWS_SECRET_ACCESS_KEY=
-AWS_REGION=us-east-1
-LOGZIO_API_KEY=
-APPLICATION_NAME=
+export CONTENTFUL_ACCESS_TOKEN=8cd3539d1e0bf1a92c9b544c194cde662a5d00e20aa85c3bf98949e3aa2650c5
```
christiancrawford: hola, did you mean to modify this file? Feels like maybe this was meant to be a local change only?

OrdersToModerate.jsx
```diff
@@ -0,0 +1,278 @@
+/* eslint-disable  */
```
christiancrawford: Doesn't seem like this has to do with the style changes you made. Is this file meant to be added to this PR?

angelicawink: @christiancrawford  Hm wow I have no idea how this got in there. From a totally different project. removing

.env-sample
angelicawink: @christiancrawford definitely a mistake. Thank you!

PR #1231, opened 2019-12-06
Locations buttons

angelicawink:
"Set my site" button should stack **beneath** "Learn more" button at/around iPad screen width.

### DIFF
--- a/_assets/stylesheets/pages/_locations.scss
+++ b/_assets/stylesheets/pages/_locations.scss
@@ -34,25 +34,31 @@ crds-button {
 }
 crds-site-select {
     position: absolute;
-    bottom: 0;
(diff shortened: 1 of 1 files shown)


_assets/stylesheets/pages/_locations.scss
```diff
@@ -34,26 +34,32 @@ crds-button {
 }
 crds-site-select {
     position: absolute;
-    bottom: 0;
-    left: 125px;
+    // bottom: 0;
+    // left: 125px;
```
j-brammer: Kill of commented styles if you don't really need them anymore

angelicawink: thanks for the catch! removed

PR #1269, opened 2019-12-18
fix homepage, infrastructure, live and livestream test [development]

lynchl607:
## Problem
*Document current behavior and why it's undesirable here.*

## Solution
*Document how the behavior was fixed or enhanced here.*

### Corresponding Branch

## Testing

### DIFF
--- a/cypress/integration/404/404_spec.js
+++ b/cypress/integration/404/404_spec.js
@@ -14,8 +14,9 @@ describe('Testing the 404 page:', function () {
   });
 
   it('/search page should load when the search button is clicked', function () {
-    cy.get('[data-automation-id="404-search-button"]').as('404SearchButton');
-    cy.get('@404SearchButton').click();
+      cy.wait(5000);
+      cy.get('[data-automation-id="404-search-button"]').as('404SearchButton');
+      cy.get('@404SearchButton').click();
 
     cy.get('.ais-SearchBox-input').as('searchField');
     cy.get('@searchField').should('exist').and('be.visible');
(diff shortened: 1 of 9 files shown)


cypress/integration/homepage/home_page_current_series_spec.js
```diff
@@ -9,7 +9,10 @@ describe('Testing the Current Series on the Homepage:', function () {
       currentSeries = series;
     });
 
-    cy.ignorePropertyUndefinedTypeError();
+      cy.ignorePropertyUndefinedTypeError();
+      cy.on('uncaught:exception', (err, runnable) => {
+          return false
+      })
```
j-brammer: @lynchl607 I'm seeing this particular block of code several times. Is this a fix for some odd behavior?

cypress/integration/404/404_spec.js
```diff
@@ -14,8 +14,9 @@ describe('Testing the 404 page:', function () {
   });
 
   it('/search page should load when the search button is clicked', function () {
-    cy.get('[data-automation-id="404-search-button"]').as('404SearchButton');
-    cy.get('@404SearchButton').click();
+      cy.wait(5000);
```
j-brammer: I'm not a huge fan of hard coded waits unless it's completely unavoidable. I would wonder if the right thing to do is do a timeout that waits for something to be on the page and then continue. 

Something like this:
https://stackoverflow.com/questions/16149431/make-function-wait-until-element-exists/47776379 (look at green checked solution)

PR #1271, opened 2019-12-18
De7330

christiancrawford:


_assets/javascripts/lib/set-redirect-url.js
```diff
@@ -1,10 +1,21 @@
...
-    document.cookie = "redirectUrl=" + encodeURIComponent(window.location.href) + ";domain=.crossroads.net;path=/";
+    var isRedirectUrlSet = new Event("redirect-url-set");
+    var cookieName = "redirectUrl=";
+    var domainValue = document.domain;
+    var cookieValue = document.cookie
+      .split(";")
+      .find(selectedCookie => selectedCookie.startsWith(" " + cookieName));
+    var isCorrectName = cookieValue ? cookieValue.includes(domainValue) : false;
+
+    if (isCorrectName == true || cookieValue == undefined) {
```
tcmacdonald: Just so I'm understanding correctly... you are evaluating `cookieValue == undefined` so as to not overwrite a previously set value?

tcmacdonald:
Another thought @christiancrawford ... what if, instead of setting the domain to `.crossroads.net` you include the subdomain so the redirect cookie is only ever available to the environment in which it is relevant?

PR #1276, opened 2019-12-20
DE7330 | Christian | tie cookie domain to sub domain

christiancrawford:


j-brammer:
> @christiancrawford Seems like you could just omit `;domain=...` from your cookie entirely and you wouldn't have to do anything? Maybe I'm missing something.

The issue is that when we use .crossroads.net the cookies get used on int/demo/prod because the subdomain piece is effectively a wild card. Production users may never see this, but it's been causing more and more issues with those of us that move between enviroments to test things. Even if we move to Netlify controlled domains it would still set a general cookie prod would read. No bueno.

_assets/javascripts/lib/set-redirect-url.js
```diff
@@ -8,12 +8,23 @@ window.onload = function() {
...
+        subDomain +
+        ".crossroads.net;path=/";
+    }
+
+    function filterForCorrectSubdomain(subDomainPrefix) {
+      if (subDomainPrefix == "") {
+        return "www";
+      } else if (subDomainPrefix == "development") {
+        return "local";
+      }
```
j-brammer: I'm a fan of this being in a function, but I'm also a fan of ternary operations.

`return subDomainPrefix ? subDomainPrefix : 'www'` should do the trick.

Also, we don't want to change development to local. I think we would just set `CRDS.env.environment` to `local` and it would just work. This should be configurable as is.

christiancrawford: Thanks Jason ^ I didn't want to mess with CRDS.env.environment that uses 'development' already since I don't the impact this would have across the application . . . that being said, if it is better to do that then I can refactor this for sure . . . getting Jimmy's opinion on this as well.

PR #1328, opened 2020-01-30
feature/US18916-add-canonical

J-Byron:
## Problem
- Net needs to be updated to compensate for the incoming distribution channel changes

## Solution

- Removes config queries related to canonical_host
- Adds config for distribution channels site
- Bumps jekyll-contentful to 2.1.0


_config.yml
```diff
@@ -33,14 +33,21 @@ include:
   - _pages
 
 contentful:
+  config:
+    sites: "www.crossroads.net"
```
tcmacdonald: This is not an accurate use of this configuration value. Please review the docs and let me know if you have questions. https://github.com/crdschurch/jekyll-contentful#content-distribution

PR #1330, opened 2020-01-31
removed uncaught exception error handleing from home_page_current_ser…

lynchl607:


cypress/support/index.js
```diff
@@ -16,6 +16,11 @@
 // Import commands.js using ES2015 syntax:
 import './commands';
 
+Cypress.on('uncaught:exception', (err, runnable) => {
```
s-henney: Instead of adding this here, could you create a custom command in the support/commands.js file? That's where other ignore uncaught exception commands are.

s-henney:
Also, why is this being merged into master not development?

cypress/support/index.js
lynchl607: done

cypress/integration/homepage/home_page_latest_message_spec.js
```diff
@@ -51,9 +51,11 @@ describe('Tests the Current Message on the Homepage', function () {
     });
   });
 
-  it('Checks card image and, if Bitmovin video, player exists and video autoplays', () => {
+  it('\itmovin video, player exists and video autoplays', () => {
```
s-henney: Not sure this deletion was intentional...

lynchl607: done

PR #1355, opened 2020-02-14
add erroe handling to home_page_latest_messgae

lynchl607:

### DIFF
--- a/cypress/integration/homepage/home_page_latest_message_spec.js
+++ b/cypress/integration/homepage/home_page_latest_message_spec.js
@@ -26,13 +26,10 @@ describe('Tests the Current Message on the Homepage', function () {
     });
 
     //Navigate
-      cy.ignorePropertyUndefinedTypeError();
-      cy.on('uncaught:exception', (err, runnable) => {
-          return false
-      })
-
-    cy.visit('/');
-  });
+  const errorsToIgnore = [/.*Cannot set property\W+\w+\W+of undefined.*/, /.*Cannot set property staus or undefined.*/];
+  cy.ignoreMatchingErrors(errorsToIgnore); 
+  cy.visit('/');
+ });
 
   it('Checks title, image, and "View latest now" button have correct link', () => {
     currentMessage.getURL().then(url => {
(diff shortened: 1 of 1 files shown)


cypress/integration/homepage/home_page_latest_message_spec.js
```diff
@@ -52,6 +52,8 @@ describe('Tests the Current Message on the Homepage', function () {
   });
 
   it('Checks card image and, if Bitmovin video, player exists and video autoplays', () => {
+  const errorsToIgnore = [/.*Cannot set property\W+\w+\W+of undefined.*/, /.*Cannot set property staus or undefined.*/];
+  cy.ignoreMatchingErrors(errorsToIgnore);
```
s-henney: Looks like you have conflicting error message handling in this file. When you want Cypress to ignore an exception, you need to overwrite how it handles the 'uncaught:exception' event. In this file alone you're doing this 3 times: 2 times in the Before clause with "cy.ignorePropertyUndefinedTypeError();" and "cy.on('uncaught:exception', (err, runnable) => { return false })", and one time here with "cy.ignoreMatchingErrors(errorsToIgnore);". These handlers are not like catch clauses - they don't stack, so only one is really active at a time, and because of how Cypress executes actions I don't know if it's guaranteed which is active when, and if they are properly scoped to a test. TL;DR - Only one way to handle the 'uncaught:exception' event should be used in any file.
Here's what I'd recommend: The 'uncaught:exception' handle should be setup before the first cy.visit() call in the Before() clause since both tests are running off the same page load. You need to handle multiple errors so I'd use the cy.ignoreMatchingErrors() helper function like you've done in lines 55-56, but move those lines up to the Before clause to replace lines 29-32. You'll want to grab the error "cy.ignorePropertyUndefinedTypeError()" and add that to the list too.
Side note, you really really really don't want to use "cy.on('uncaught:exception', (err, runnable) => { return false })" since that'll ignore every error and undermine the ability of Cypress to catch errors happening behind the scenes that often go unnoticed.

lynchl607: done

PR #1383, opened 2020-03-09
US19097 Update Location Schema

silasbaur:


_layouts/location.html
```diff
@@ -381,4 +381,8 @@ <h4 class="card-title card-title--overlap text-uppercase">
   {% include _fb-pixel.html %}
 {% endif %}
 
+<script type="application/ld+json">
+  {{ page.schema | jsonify }}
+</script>
```
tcmacdonald: Can you wrap this in an if statement so it only outputs this script tag if `page.schema` exists?

PR #1393, opened 2020-03-12
US18996 OSG Location Pages

tcmacdonald:


lib/crds/redirects.rb
```diff
@@ -37,12 +37,14 @@ def auth_required
     pages + logins
   end
 
-  def to_csv!(path = './redirects.csv')
+  def to_csv!(path = './redirects.csv', debug=true)
```
dholloran-ample: Was this `debug=true` intentionally left?

tcmacdonald: Yes, I added this flag so I could reduce output in the test suite. Something about the colorized string utility was throwing goofy issues when running the test locally so this seemed like a sensible addition.

dholloran-ample: Makes sense

PR #1407, opened 2020-03-18
Feature/us19035 remove group detail page

dholloran-ample:
This PR removes the old Onsite Groups Detail page and 301 redirects them back to `/groups/onsite` to remove 404 issues

- [Rally](https://rally1.rallydev.com/#/66096747656d/custom/24109743995?qdp=%2Fdetail%2Fuserstory%2F370828221712)


redirects.csv
```diff
@@ -135,3 +135,4 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /healing/divorcecare,/groups/onsite-healing-groups/divorce-care,301!
 /healing/postpartum-healing,/groups/onsite-healing-groups/postpartum,301!
 /*,/404.html,404
+/groups/onsite-groups/*,/groups/onsite,301!
```
tcmacdonald: @dholloran-ample This rule won't ever be evaluated b/c it falls after the wildcard rule. Can you move it up by one line?

PR #1425, opened 2020-03-25
DE7599 Netlify Redirects Development

silasbaur:
## Problem
On April 7th, there is an update to how Netlify handles redirects. Redirects will no longer be assumed to be forced so we need to add "!" to make all redirect forced in every project.

## Solution
add "!" to redirects


silasbaur:
Screen Shot 2020-04-01 at 11 00 04 AM
@tcmacdonald The redirects in contentful should be fine as long as they were inputed into contentful with the correct "force redirect?" value.

tcmacdonald:
Ok, thanks @silasbaur!

PR #1433, opened 2020-03-30
[prod] Deploy Giving Redirect

tcmacdonald:


_functions/env-redirector.js
```diff
@@ -0,0 +1,26 @@
...
+
+      }
+    }
+  } catch (err) {}
+
+  return callback(null, {
+    statusCode: 307,
+    headers: { Location: dest }
+  });
+};
```
j-brammer: I might have the syntax wrong for the function, but you get the idea. I corrected the referrer misspelling.

Goal is to remove the unnecessary const and break out the functionality into a smaller function. Also removed the dest declaration and just returned the string itself since it's your default case.

```suggestion
'use strict';

exports.handler = (event, context, callback) => {


  return callback(null, {
    statusCode: 307,
    headers: { Location: this.getDestination(event.headers.referer) }
  });
};

function getDestination(referrer) {
  try {
    if(referrer) {

      // if referrer resembles INT or local address...
      if(['development', 'localhost', '0.0.0.0', '127.0.0.1', 'int' ].some(el => referrer.includes(el))) {
        return "https://sandbox.pushpay.io/g/crossroadscincinnatiint"

      // if referrer resembles DEMO address...
      } else if (['demo', 'release'].some(el => referrer.includes(el))) {
        return "https://sandbox.pushpay.io/g/crossroadscincinnatilegacy"

      }
      return "https://pushpay.com/g/crossroads";
    }
  } catch (err) {}
}
```

PR #1452, opened 2020-04-15
DE7628 Redirect Url Logic

silasbaur:
## Problem
[Rally Defect](https://rally1.rallydev.com/#/38108142417d/detail/defect/385257475596)

## Corresponding PR
[Crds-Components](https://github.com/crdschurch/crds-components/pull/465)


redirects.csv
```diff
@@ -2,7 +2,9 @@
 /group-renew groupIds=:groupIds,/group-renew?groupIds=:groupIds,200! Role=user
 /group-renew groupIds=:groupIds,/signin?redirectUrl=/group-renew?groupIds=:groupIds,302!
 http://crossroads.net/*,https://www.crossroads.net/:splat,301!,master
+https://crossroads.net/*,https://www.crossroads.net/:splat,301!,master
 http://${env:CRDS_APP_DOMAIN}/*,https://${env:CRDS_APP_DOMAIN}/:splat,301!
+https://${env:CRDS_APP_DOMAIN}/*,https://${env:CRDS_APP_DOMAIN}/:splat,301!
```
tcmacdonald, line 7: Not sure what you're trying to accomplish here but this rule will create an infinite loop. 

`CRDS_APP_DOMAIN == 'www.crossroads.net'`

silasbaur: @tcmacdonald I did this to cover both http and https

PR #1535, opened 2020-06-09
US19669 Location About

silasbaur:
## Problem

- [ ] Add about this location header to template
- [ ] Add watch with us button linked to the page
- [ ] "Add social media buttons above form that intitate following on Instagram and Facebook. This should be editable in the CMS buttons should only appear if content exists and if no exist omit entirely
- [ ] Clean up content for removed content
- [ ] Update form
- [ ] Add horizontal rule and drop in additional infor copy about sign language
- [ ] Update background

### DIFF
--- a/_assets/stylesheets/components/_hubspot-forms.scss
+++ b/_assets/stylesheets/components/_hubspot-forms.scss
@@ -3,8 +3,8 @@ body .hs-button.primary,
 body input[type="submit"],
 body input[type="button"]:not(#getStartedBtn):not(#myStuffBtn) {
   color: #fff;
-  background-color: #3b6e8f;
-  border-color: #3b6e8f;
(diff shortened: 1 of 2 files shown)


_assets/stylesheets/components/_hubspot-forms.scss
```diff
@@ -3,8 +3,8 @@ body .hs-button.primary,
 body input[type="submit"],
 body input[type="button"]:not(#getStartedBtn):not(#myStuffBtn) {
   color: #fff;
-  background-color: #3b6e8f;
-  border-color: #3b6e8f;
+  background-color: #253746;
+  border-color: #253746;
```
dholloran-ample, line 7: Can we use a variable for this or add a new one? If not could you add an inline comment explaining why

PR #1562, opened 2020-06-16
US19707/Adds updates to homepage logged in

kellgauthier:
## Task
- Update links to $cr-orange
- Update timestamp background to $cr-black 
- Update image overlay to $cr-black 

[Rally](https://rally1.rallydev.com/#/66096747656ud/detail/userstory/398360211236?fdp=true)

Buttons will be updated with a border-radius of 0 on another story.

### DIFF
--- a/_assets/stylesheets/_utilities.scss
+++ b/_assets/stylesheets/_utilities.scss
@@ -25,20 +25,8 @@
   background-image: linear-gradient(
     to bottom,
     rgba(23, 23, 23, 0),
-    $cr-gray-darker
-  );
-}
-
-.bg-overlay-black {
-  position: absolute;
-  top: 25%;
-  right: 0;
-  left: 0;
-  bottom: 0;
-  background-image: linear-gradient(
-    to bottom,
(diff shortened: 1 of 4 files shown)


_assets/stylesheets/_utilities.scss
```diff
@@ -29,6 +29,19 @@
   );
 }
 
+.bg-overlay-black {
```
silasbaur, line 32: I would change the original class of `bg-overlay` instead of making a new class

_includes/media/_overlay-card.html
```diff
@@ -28,7 +28,7 @@
 
 <a href="{{ item | media_url }}" class="overlay-card{% if include.size == 'xl' %} overlay-card-xl{% endif %}">
   {% if item.bitmovin_url == null %}
-    <div class="bg-overlay"></div>
+    <div class="bg-overlay-black"></div>
```
silasbaur, line 31: Change this back to just `bg-overlay`

PR #1588, opened 2020-06-19
Statement of faith

jakecatron:


statementoffaith.html
```diff
@@ -0,0 +1,59 @@
+---
+layout: default
+---
+
+<section class="jumbotron jumbo-height" style="background-image: url('//crds-media.imgix.net//23IWaWUAICoAUG4Jassq40/24e97abb4beedcc1e41c334e06d4305e/crossroads-church-old2.png'); height: 455px;">
+    <h1 class="component-header-box component-header-box-border-lg mt-100">
```
tcmacdonald, line 6: `mt-100` is not a thing. Can you please remove? Thanks!

statementoffaith.html
```diff
@@ -0,0 +1,59 @@
+---
+layout: default
+---
+
+<section class="jumbotron jumbo-height" style="background-image: url('//crds-media.imgix.net//23IWaWUAICoAUG4Jassq40/24e97abb4beedcc1e41c334e06d4305e/crossroads-church-old2.png'); height: 455px;">
```
silasbaur, line 5: What is `jumbo-height` doing?

jakecatron: ah, thanks for catching that. It was an old class not being used anymore.

statementoffaith.html
jakecatron: Sorry about this, I should've left a note that it was on the our-history branch.

tcmacdonald:
Looks good @jakecatron – should this page have a header / footer?

PR #1605, opened 2020-06-22
DE7821 Most Recent

silasbaur:
## Problem
The "articles, podcasts and more" section of logged out homepage is supposed to show the three most recent pieces of content. Two issues I'm seeing here:

1. These aren't the three most recent content items (not even the three most recent articles)
2. One of them is duplicated (the big feature one and the lower right)
image

## Solution
[Preview](https://deploy-preview-1605--int-crds-net.netlify.app/)

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,7 +34,6 @@ paginate:
       - articles
     per: 10
     limit: 1
-    offset: 2
     sort: published_at desc
   episodes:
     per: 3
@@ -142,6 +141,7 @@ notification:
   </div>
 </section>
 
+
 {% comment %}Get Equipped: Watch the Latest Teaching{% endcomment %}
 <section class="bg-tan-light soft-bottom">
   <div class="container soft-ends hard-sides">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -213,29 +220,27 @@ <h2 class="component-header-box text-orange">Articles, Podcasts, & More</h2>
       <p class="text-blue font-family-serif font-size-large">Unfiltered & authentic takes on life</p>
     </div>
     <div class="col-sm-8">
-      {% assign below_the_fold = page.home_collection.docs | slice: 4 %}
-      {% for article in below_the_fold %}
+      {% assign most_recent = page.home_collection.docs | sort: "published_at" | reverse | first %}
```
dholloran-ample, line 223: I think the `reverse` filter is causing the newest article to be last and the 2nd newest first. Otherwise, this seems right to me

screenshot 2020-06-22 at 15 50 39

index.html
```diff
@@ -203,6 +203,13 @@ <h2 class="component-header-box text-blue">Gear Up for Your Adventure</h2>
   </div>
 </section>
 
+<section>
```
dholloran-ample, line 206: I think this `<section/>` needs to be removed looks like something you were testing?

screenshot 2020-06-22 at 15 53 46

index.html
tcmacdonald: Thanks @dholloran-ample. I think the discrepancy is that you're sorting via `updated_at` while he's sorting via `published_at`.

PR #1657, opened 2020-07-01
US19799 Best Content CMS

silasbaur:
## Problem
Enable content updates for Our Best Content (featured media) section on logged out home via CMS. Use the featured media content type

## Solution
[Preview](https://5f04ac4bfaedeb0008fad478--int-crds-net.netlify.app/)


kellgauthier:
@silasbaur - on the deploy preview and when I spin up locally, I see this. 
Screen Shot 2020-07-07 at 12 13 51 PM

PR #1848, opened 2020-09-30
setup the redirects for event check in

candrews55:


netlify.toml
```diff
@@ -28,7 +28,7 @@ package = "./_plugins/netlify-plugin-health-check"
   CRDS_ANGULAR_ENDPOINT = "crds-angular-int.netlify.app"
   CRDS_SERVE_ENDPOINT = "sign-up-to-serve-int.crossroads.net"
   CRDS_PROFILE_ENDPOINT = "profile-int.crossroads.net"
-  CRDS_EVENT_CHECKIN_ENDPOINT = "development.event-checkin.crossroads.net"
+  CRDS_EVENT_CHECKIN_ENDPOINT = "eventcheckin-int.crossroads.net"
```
briancavanaugh, line 31: To be consistent, should the changes in this file be "event-checkin" instead of "eventcheckin"

candrews55: That would cause problems with redirecting people from the old sub domain url to crds.net/event-checkin in prod.

briancavanaugh (approved):

PR #1878, opened 2020-10-14
update index on location and onsite group test

lynchl607:
update index on location and onsite group test

### DIFF
--- a/cypress/integration/locations/locations_landing_page_spec.js
+++ b/cypress/integration/locations/locations_landing_page_spec.js
@@ -29,7 +29,7 @@ describe('Given I navigate to /locations and do not search:', function() {
     let location;
     let name;
     before(function() {
-      location = locationList[index - 0];
+      location = locationList[index];
       name = location.name.text;
     });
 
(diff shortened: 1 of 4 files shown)


cypress/integration/locations/locations_landing_page_spec.js
```diff
@@ -21,15 +21,15 @@ describe('Given I navigate to /locations and do not search:', function() {
...
+      .find('[data-automation-id="location-name"]')
+      .should('have.attr', 'href', '/anywhere');
   });
 
   [1, 2, 3].forEach((index) => {
     let location;
     let name;
     before(function() {
-      location = locationList[index - 1];
+      location = locationList[index - 0];
```
s-henney, line 32: Alternatively `locationList[index]`

cypress/integration/onsite/onsite_group_landing_page_spec.js
```diff
@@ -36,16 +36,16 @@ describe('Given I navigate to /onsite/group Page:', function() {
...
+      .first()
       .find('a')
       .should('have.attr', 'href', '/groups/onsite/financial-peace/uptown');
   });
 
-  [0, 1, 2].forEach((index) => {    
+  [1,2,3,].forEach((index) => {    
     it(`Onsite card #${index} should have a Title`, function() {
-      let title = onsiteGroupList[index].title.text;
+      let title = onsiteGroupList[index + 8].title.text;
```
s-henney, line 48: Reporting that we're testing card at index 1 when we're actually testing the card at index 9 is very misleading. Can you change the array from 1-3 to 9-11 if those are the indexes of the cards that should be tested?

s-henney (changes requested):

s-henney (approved):

PR #1885, opened 2020-10-22
Prod Deploy 10/22/20

silasbaur:


redirects.csv
```diff
@@ -47,7 +46,7 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /childcare,https://${env:CRDS_ANGULAR_ENDPOINT}/childcare,200!
 /childcare/*,https://${env:CRDS_ANGULAR_ENDPOINT}/childcare/:splat,200!
 /corkboard*,https://${env:CRDS_ANGULAR_ENDPOINT}/corkboard:splat,200!
-/invoices/*,https://${env:CRDS_ANGULAR_ENDPOINT}/invoices:splat,200!
+/invoices/*,https://${env:CRDS_INVOICES}/:splat,200!
```
dholloran-ample, line 49: This environment variable is not set in any of the environments.

dholloran-ample (changes requested):
@silasbaur - There is a missing env variable (see inline comment) that is set in the `redirects.csv`. We need to either need to set the env or remove the commit(s)

dholloran-ample (approved):

PR #1914, opened 2020-11-18
Prod Deploy 11-18-20

silasbaur:


candrews55 (approved):

redirects.csv
```diff
@@ -53,7 +53,6 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /profile/personal,/profile/personal-info,302!
 /profile/*,https://${env:CRDS_PROFILE_ENDPOINT}/:splat,200! Role=user
 /profile/*,/signin?redirectUrl=/profile/:splat,302!
-/signin,https://${env:CRDS_ANGULAR_ENDPOINT}/signin,200!
```
tcmacdonald, line 56: @j-brammer It looks like you're removing responsibility from crds-angular-legacy which is awesome but now we can auth concerns in two places. Can you talk me through your strategy / reasons for this move?

j-brammer: Redirects added back in after discussion with Ample

tcmacdonald: Thanks @j-brammer !

tcmacdonald (approved):

PR #1967, opened 2020-12-21
Onsite Groups Landing Page

tcmacdonald:
[Preview](https://deploy-preview-1967--int-crds-net.netlify.app/groups/site-based/)

https://projects.invisionapp.com/d/main#/console/20902251/440281252/preview

- [ ] 'Back' link on detail page should reflect group type, link should work.
- [ ] How do we handle the redirect for /groups/onsite?
- [ ] Should the 'anywhere' location still be 'Online | Crossroads Anywhere'?
- [ ] Mobile tabs don't scale, should we handle this in a separate story later?
- [ ] Topo map on tabs is impossible to build unless somebody has some tricks up their sleeve?
- [ ] Added a new content-block liquid tag that should be ported `jekyll-crds`
- [ ] Content blocks are not parsing markdown


beckysingson:
@tcmacdonald Leaving comments here since this work doesn't have a corresponding Rally story at the moment:

- This new category-based page structure creates an issue on the meeting detail pages - the convention breaks the "Back to All Onsite Groups" link ([example](https://deploy-preview-1967--int-crds-net.netlify.app/groups/onsite/basic-bible-onsite-group/lexington/)). I might suggest we just remove that unless you want to infer the URL based on the category to which the meeting is assigned? Thoughts?

Crossroads

-  Oy, these tabs on mobile be hurting my soul.
- Unrelated to this body of work but leaving the comment here so I don't forget to say it - RE: Osano: that lil nom nom cookie icon in the corner is going to interfere with the Intercom chat bubble. Does Osano allow us to toggle that off and/or place it elsewhere?

silasbaur (approved):

PR #1999, opened 2021-01-11
replace app image

jakecatron:


adamtgeorge (approved):
@jakecatron  - looks great

PR #2014, opened 2021-01-20
Prod Deploy 1-20-21

silasbaur:


candrews55 (approved):

dholloran-ample (approved):

_assets/javascripts/components/resi-player.js
```diff
@@ -0,0 +1,59 @@
+const isDST = (d) => {
```
mjvezzani, line 1: I see we are creating custom JavaScript time utility methods. I don't have an issue with this, but was wondering if it makes sense at all to use something like the JS library Moment and Moment Timezone. That way we can offload the responsibility for getting time right on a more robust library.

mjvezzani (approved):
I'm not going to block this because of my comment on the `resi-player.js` file. I was largely curious to know if consideration had been given to using something like Moment.js.

j-brammer (approved):

PR #2074, opened 2021-03-02
Feature/us20697 netlify training proof of concept

mjvezzani:
## Problem
Training team needs a way to show uploaded trainings from the Rise LMS system that have been deployed to Netlify.

## Solution
Update the redirects.csv file to provide a redirect that shows the training that is hosted on Netlify

## Testing
Visit www.crossroads.net/training/listening and confirm the listening training is visible and interactive.


mjvezzani:
I recognize that these new redirect rules will need changing next week, but this does get the PoC (which was the ask for this week) to work.

Once this work has been merged, visit https://int.crossroads.net/training/listening to view the training.

PR #2078, opened 2021-03-04
DEXXXX Shows Redirect

silasbaur:
## Problem
When going to crossroads.net/media/shows/ you sometimes get redirected to crossroads.net/shows/ which is a 404. I can recreate this issue 2/10 time in an incognito window

## Solution
Add a redirect for `/shows` to goto `/media/shows` for the small percentage of times it is redirected to `/shows`


tcmacdonald:
Closed in lieu of #2106

PR #2133, opened 2021-03-30
Hotfix richmond remove office loc

chriseallen:
## Problem
Don't want Richmond's Office Address on page

## Solution
Removed it by commenting out for now incase they want it back.


silasbaur (changes requested):
@chriseallen Lets make sure you only have your changes and commits in this PR. There are 11 files changed when there really should be just 1

chriseallen:
@silasbaur Do you need this is because I pulled the dev branch? I want to make sure those commits don't somehow sneak into the file somehow. I really don't know how they got there.

silasbaur:
@chriseallen Yes that is how that happened. If you need to make a PR for master, you should branch off of master to make sure we do not have anything extra in your PRs

PR #2162, opened 2021-04-07
Us20885 established jumbotron

jakecatron:
Jumbotron for established locations.


jakecatron:
@silasbaur Hey Silas - the established jumbotron is good to go with logic and design but there were a couple small issues I'm having that I don't want to spend anymore time on until you're back in the office:
1. There is horizontal scroll on the page that I cannot find the root cause of. I'm not sure when it was introduced but I have tried to narrow it down to certain sections of the page and I just can't figure it out.
2. The pastor image gets distorted on different screen sizes due to flexbox. I'm not sure if there's any easy solution for this or if we need to add it to its own column to resolve the issues but I just don't want to waste anymore time on that one right now.

Thanks man, hope you enjoyed your time off!

PR #2166, opened 2021-04-09
Us20885 established jumbotron

jakecatron:
Established jumbotron for locations redesign.


jakecatron:
@silasbaur Hey Silas - the established jumbotron is good to go with logic and design but there were a couple small issues I'm having that I don't want to spend anymore time on until you're back in the office:

There is horizontal scroll on the page that I cannot find the root cause of. I'm not sure when it was introduced but I have tried to narrow it down to certain sections of the page and I just can't figure it out.
The pastor image gets distorted on different screen sizes due to flexbox. I'm not sure if there's any easy solution for this or if we need to add it to its own column to resolve the issues but I just don't want to waste anymore time on that one right now.
Thanks man, hope you enjoyed your time off!

_layouts/established.html
```diff
@@ -0,0 +1,118 @@
...
+layout: default
+---
+
+<!-- Pull only the promos that exist for the specific location if any -->
+{% assign location_happenings = site.promos | where: "target_audience", page.name | sort: "date" | reverse %}
+{% assign series = site.series | last %}
+{% include header.html %}
+
+<!-- Resi Player -->
+{% assign locations_with_live_stream =  "oakley, florence, mason, eastside, westside, lexington, uptown" | split: ", "%}
```
dholloran-ample, line 11: Let's move Resi into a partial so we do not need to maintain 2 sets of live stream locations

_layouts/established.html
```diff
@@ -0,0 +1,118 @@
...
+</section>
+
+<!-- Featured Section - Below Jumbotron -->
+{% if page.featured_below.size > 0 %}
+  {{ page.featured_below }}
+{% endif %}
+
+
+
+<!-- Modal -->
```
dholloran-ample, line 72: Is there a reason why we are not using `_includes/_subscribe-modal.html`? It looks pretty similar but I could be missing something.

dholloran-ample (changes requested):
@silasbaur cc: @jakecatron I left 2 inline comments.

_layouts/established.html
silasbaur: The hubspot form is different for each location

dholloran-ample (approved):

PR #2191, opened 2021-04-26
US20989 Humans Pagination

silasbaur:
## Task
Implement pagination on the humans of crossroads index page

## Solution
[Preview](https://humans-pagination--demo-crds-net.netlify.app/media/humans-of-crossroads/) for testing. (added some more articles so that we can test if pagination works)

### DIFF
--- a/_assets/stylesheets/_utilities.scss
+++ b/_assets/stylesheets/_utilities.scss
@@ -62,6 +62,10 @@ a.hover-text-gray-lighter:hover * {
   background-color: $cr-black;
 }
 
+.bg-gray-darkest {
+  background-color: $cr-gray-darkest !important;
+}
+
 .border-white-15 {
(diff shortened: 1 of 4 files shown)


dholloran-ample (changes requested):
@silasbaur It seems as it works for me I have 2 improvements and 2 stylistic changes.

## Improvements
1. Could we use push-state to update the URL.
2. when I go to a page other than the first page `/media/humans-of-crossroads/page/3/` Could we do one of the following
    1. Load the previous pages?
    2. Fallback to a numbered pagination?
    3. Not show the button?

## Style issues
1. Load more button overlaps the footer.
screenshot 2021-04-26 at 12 14 18
2. Spacing between new pagination group is different ~77px() than the other items (~54px
screenshot 2021-04-26 at 12 10
)

dholloran-ample (approved):

PR #2212, opened 2021-05-06
My impact v1

dtkocher:
## Problem
Need to capture when a video or message is watched for Personal Engagement feature.

## Solution
Render the crds-media-metric component and fire off an event that the component is listening for.

## Testing
Watch a video  and then go to azure cosmos db and see that a media metrics collection has a record for that video

### DIFF
--- a/_assets/javascripts/components/bitmovin.js
+++ b/_assets/javascripts/components/bitmovin.js
@@ -204,8 +204,8 @@ class BitmovinManager {
   }
 
   onPlayerStart() {
-    var defJSLoaded = new Event('mediaMetricInteraction');
-    document.dispatchEvent(defJSLoaded);
+    const mediaMetric = new Event('mediaMetricInteraction');
+    document.dispatchEvent(mediaMetric);
 
     if (this.getIsMuted()) this.enableSubtitles();
     if (typeof analytics !== "undefined") {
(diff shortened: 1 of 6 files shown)


silasbaur (approved):

_assets/javascripts/components/bitmovin.js
```diff
@@ -204,6 +204,9 @@ class BitmovinManager {
   }
 
   onPlayerStart() {
+    var defJSLoaded = new Event('mediaMetricInteraction');
+    document.dispatchEvent(defJSLoaded);
```
janssejl, line 208: you migh want to rename this var. I know this is from when I sent you an example. defJSLoaded is referencing "deferred javascript file loaded".

_assets/javascripts/components/youtube.js
```diff
@@ -74,6 +74,11 @@ class YoutubeManager {
         });
       }
     }
+
+    if (event.data == YT.PlayerState.PLAYING) {
+      var defJSLoaded = new Event('mediaMetricInteraction');
```
janssejl, line 79: same here with the var name.

janssejl (changes requested):
Nice with the docker set up. Everything looks great other than those var names

PR #2214, opened 2021-05-06
Add event to fire off media metric for article views after 30 seconds

dtkocher:
## Problem
Track article views

## Solution
Call crds-media-metrics component after 30 seconds to send off an update that the article had been viewed.


JoshNieman (approved):
lgtm

PR #2223, opened 2021-05-07
Details don't always have appropriate spacing between columns

kellgauthier:
## Problem
[Asana](https://app.asana.com/0/1198951831278010/1200290763427528)

### DIFF
--- a/_layouts/established.html
+++ b/_layouts/established.html
@@ -175,7 +175,7 @@ layout: default
                     text="More about Student Ministry">
                   </crds-button>
                 </div>
-                <div class="df-screen-sm">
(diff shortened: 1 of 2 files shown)


silasbaur (changes requested):
Make sure you check every location. This is the Florence location on int
Screen Shot 2021-05-07 at 1 26 21 PM

kellgauthier:
Hi @silasbaur - Thanks for the feedback. I think at the location level it becomes a content administration problem and once merged I can fix in the locations content model in contentful. See screenshot: 
Screen Shot 2021-05-07 at 2 04 46 PM

silasbaur (approved):

PR #2232, opened 2021-05-10
Update spacing on Jumbotron between headers and content and links

kellgauthier:
[Asana](https://app.asana.com/0/1199580851373247/1200290763427495)


_assets/stylesheets/pages/_locations.scss
```diff
@@ -17,6 +17,10 @@
     }
   }
 
+  p {
```
silasbaur, line 20: This will affect every p tag on the site

PR #2280, opened 2021-06-01
US20386 Article Redesign

silasbaur:
## Task
[Design](https://invis.io/HNYKN2CF7VJ#/436889269_Article_V2_A)

## Solution
[Deploy Preview](https://article-redesign--int-crds-net.netlify.app/media/articles/handling-depression-how-to-keep-getting-up)

## Corresponding PR
[Crds-Component (heart-button)](https://github.com/crdschurch/crds-components/pull/1261)

### Notes
Changes I made that contradict the design
- I kept the media object component the same. Only difference was that the card stamp was white instead of black
- In design there are 7 authors shown in the "Popular Author" section but I changed it to show 6 because 7 does not work with our 12 column layout.

### DIFF
--- a/_layouts/article.html
+++ b/_layouts/article.html
@@ -37,6 +37,10 @@ title: Articles
 
         <!-- Content -->
         <div class="soft-ends article-tpl-content font-size-base">
+          {% if page.lead_text %}
+          <div class="lead">{{ page.lead_text | markdownify }}</div>
+          {% endif %}
+          
           {{ page.content }}
         </div>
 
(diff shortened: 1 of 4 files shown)


_layouts/article.html
```diff
@@ -1,220 +1,244 @@
...
+        </div>
+        <div class="container text-center">
+          <p class="font-family-serif push-top soft-top text-white">Subscribe to get our best content delivered straight to your inbox.</p>
+        </div> 
+      </div>
+    </div>
+  </section>
 </article>
-
-<script type="text/javascript">
```
dholloran-ample, line 213: Was this `mediaMetricView` intentionally script removed?

dholloran-ample (changes requested):
@silasbaur - I have a few things (told you I would)

1. The body copy looks smaller than the design (if this is intentional let me know).
1. Add more spacing between **Meet the author** and **Popular In ...**
1. Can we use the cards from `/media/articles/` in the **Popular In ...** section to match the design (closer)?
1. For some reason your new heart button style is not showing (I checked the theme prop and the deployment completed everything looks right not sure what is missing)
1. 1 inline comment/question

_layouts/article.html
silasbaur: Nope I will add it back

silasbaur:
@dholloran-ample 
1. I can fix that
2. I can fix that
3. I decided to use these cards instead so we do not keep running into this issue
Screen Shot 2021-06-02 at 3 18 19 PM
4. I just updated the preview. Should show up now

dholloran-ample (changes requested):
@silasbaur - I'm good with the changes you made and the reasoning for the cards. This may be a nitpick but the headers seem a little large maybe we should bring back this [styling](https://github.com/crdschurch/crds-net/pull/2280/files#diff-1d4a67eb4c8c4035f249a2cce973b7e1b4e421e20c77a099481ff8128b99afd1L51-L60) or something like it. Maybe instead of `1.5rem`  maybe `2.0rem`.

dholloran-ample (approved):

dtkocher (approved):

PR #2316, opened 2021-06-17
Remove Site Nav from Trips

adamtgeorge:


silasbaur (commented):
Any reason for getting rid of the front matter? I would make this a layout of `no-header-no-footer` 
I think this error will go away once it is merged and not on a deploy preview url
Screen Shot 2021-06-17 at 12 08 29 PM

silasbaur (approved):

PR #2321, opened 2021-06-21
Defect/us21210 view all happenings 

laurasprauer:
[Rally Task](https://rally1.rallydev.com/#/66096747656d/teamboard?detail=%2Fuserstory%2F600916034915)

Must be released with this PR: https://github.com/crdschurch/crds-components/pull/1289


dholloran-ample (approved):

sarahshiffman:
@laurasprauer I added promos to Oakley on INT and I see where you added View All, but it still shows all to begin with and that button doesn't expand anything? We should only show 5 and then have the View All option to expand the additional promos.  https://int.crossroads.net/oakley
image

PR #2333, opened 2021-06-28
Remove onPostBuild hook to rename ga references

mjvezzani:
## Problem
Our previous implementation of gtag.js was resulting in double reporting to google analytics

## Solution
This code removes the code that renamed all the references to `ga()` to `gtag()`.

### Corresponding Branch
No corresponding branches.

## Testing
This will be done manually by inspecting realtime traffic to our google analytics instance.

### DIFF
--- a/_includes/_head.html
+++ b/_includes/_head.html
@@ -103,15 +103,6 @@
   <script>
   var CRDS = window.CRDS || {};
 
-  // GOOGLE ANALYTICS
-  window.dataLayer = window.dataLayer || [];
-  function gtag(){dataLayer.push(arguments);}
-  gtag('js', new Date());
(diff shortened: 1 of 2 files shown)


dholloran-ample (changes requested):
@mjvezzani Should we also remove gtag.js from [`_head.html`](https://github.com/crdschurch/crds-net/blob/ddb7bde8d6f707e3fe6e1243e5dea1f26c58a502/_includes/_head.html#L106-L113) as well since it is being loaded in GTM (Media INT GA4 Configuration)? It does not look like this is needed anymore but I could be missing something.

dholloran-ample (approved):

PR #2335, opened 2021-06-29
US20611 Hide Jumbotron Videos on Mobile - pt 2

MariaSzubski:
## Problem
PR #2329 hides the video container on mobile but does not prevent video from loading in background. It also hide the inline-video and controls after user clicks the in-jumbotron play button.

## Solution
- Remove `close-video` and `inline-video-player` from `display: none` list
- Handle video loading in javascript:
    - the video `src=""` attribute is now `data-src=""` by default and won't load the background video
    - An event listener tracks the 480px media query
    - When the page is above 480px, `data-source=""` changes to `src=""` and the video loads
    - `src=""` changes back to `data-src=""` when the page crosses below 480px and the video unloads

### Corresponding Branch
PR #2329 which was previously merged into `development`.

## Testing
Test the home page below and above 480px wide. 
- Starting below 480px, the source file `spiritual-outfitters-loop-comp.mp4` won't show up in the network tab in the inspector until the page is resized above 480px.
- Resizing back below 480px will unload the video and it will re-start from the beginning the next time the window is resized above mobile.

### DIFF
--- a/_assets/javascripts/components/jumbotron-video.js
+++ b/_assets/javascripts/components/jumbotron-video.js
@@ -1,5 +1,24 @@
 (function() {
   window.$(document).ready(function() {
     new window.CRDS.JumbotronVideos();
+
+    // Prevent background video from loading on mobile
+    const mediaQuery = window.matchMedia('(min-width: 480px)');
+    mediaQuery.addEventListener('change', handleScreenChange);
+
+    // Check screen size on load
+    handleScreenChange(mediaQuery);
+
+    function handleScreenChange(e) {
+      var bgVideo = document.getElementById('jumbotron-bg-video');
+
+      if (e.matches) {
+        bgVideo.setAttribute('src', bgVideo.getAttribute('data-src'));
+      } else {
+        if (!bgVideo.hasAttribute('data-src')) {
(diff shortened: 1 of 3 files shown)


_assets/javascripts/components/jumbotron-video.js
```diff
@@ -1,5 +1,24 @@
...
+    function handleScreenChange(e) {
+      var bgVideo = document.getElementById('jumbotron-bg-video');
+
+      if (e.matches) {
+        bgVideo.setAttribute('src', bgVideo.getAttribute('data-src'));
+      } else {
+        if (!bgVideo.hasAttribute('data-src')) {
+          bgVideo.setAttribute('data-src', bgVideo.getAttribute('src'));
+        }
+      }
```
silasbaur, line 21: Getting errors on pages without the jumbotron video
Screen Shot 2021-06-30 at 2 07 16 PM

silasbaur (changes requested):

_assets/javascripts/components/jumbotron-video.js
MariaSzubski: @silasbaur Fixed!

silasbaur (approved):

PR #2350, opened 2021-07-08
Article Index Pagination

mjvezzani:
Pagination on the article index is now in the format of [n] of NN, where
"n" is the current page your are on, "[]" is the input field that allows
you to change which page you are on, and "NN" is the total number of
paginated pages.

## Problem
It was requested that we do a more traditional pagination solution for
the articles index page.

## Solution
This PR adds in code to provide a more traditional pagination experience.

### Corresponding Branch
No corresponding branches.

## Testing
Visit /media/articles/ and use the navigation provided at the bottom of the page.

### DIFF
--- a/_assets/javascripts/components/pagination.js
+++ b/_assets/javascripts/components/pagination.js
@@ -83,4 +83,53 @@ CRDS.PageLoader.prototype.activatePage = function(type) {
 
 $(document).ready(function() {
   new CRDS.Pagination();
+
+  const currentPage = parseInt(paginationInput.getAttribute("placeholder"));
+  const host = window.location.host;
+  const pageNumber = parseInt(paginationInput.value);
+  const paginationInput = document.getElementById("pagination-current-page");
+  const paginationNextButton = document.getElementById("pagination-next-button");
+  const paginationPrevButton = document.getElementById("pagination-previous-button");
+  const paginationSubmitButton = document.getElementById("pagination-submit-button");
+  const protocol = window.location.protocol;
+  const protocolAndHost = `${protocol}//${host}`
+  const totalPaginatedPages = parseInt(document.getElementById("total_pages").innerText);
+
+  paginationSubmitButton.addEventListener("click", function(e) {
+    if (pageNumber > 1 && pageNumber <= totalPaginatedPages) {
+        window.location.href = `${protocolAndHost}/media/articles/page/${paginationInput.value}/`;
+    } else if (pageNumber == 1) {
+        window.location.href = `${protocolAndHost}/media/articles/`;
+    } else {
+      console.log('Cant navigate past bounds of paginated collection');
+    }
+  }, false);
+
+  paginationInput.addEventListener("keyup", function(e) {
+    if (e.key == 'Enter') {
+      if (pageNumber > 1 && pageNumber <= totalPaginatedPages) {
+          window.location.href = `${protocolAndHost}/media/articles/page/${paginationInput.value}/`;
+      } else if (pageNumber == 1) {
+          window.location.href = `${protocolAndHost}/media/articles/`;
+      } else {
+        console.log('Cant navigate past bounds of paginated collection');
+      }
+    }
+  }, false);
+
+  paginationPrevButton.addEventListener("click", function(e) {
+    if (currentPage > 2) {
+      window.location.href = `${protocolAndHost}/media/articles/page/${currentPage - 1}/`;
+    } else {
+      window.location.href = `${protocolAndHost}/media/articles/`;
+    }
+  }, false);
+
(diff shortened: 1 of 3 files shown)


_assets/javascripts/components/pagination.js
```diff
@@ -83,4 +83,53 @@ CRDS.PageLoader.prototype.activatePage = function(type) {
 
 $(document).ready(function() {
   new CRDS.Pagination();
-});
\ No newline at end of file
+
+  const currentPage = parseInt(paginationInput.getAttribute("placeholder"));
```
dholloran-ample, line 87: I'm getting a `Uncaught TypeError: Cannot read property 'getAttribute' of undefined` error I believe is related to `paginationInput` being undefined.

How I would handle something like as follows.
```javascript
$(document).ready(function() {
  const paginationInput = document.querySelector('.pagination-selector');

  if (!paginationInput) {
    return;
  }

  new CRDS.Pagination();

  const currentPage = parseInt(paginationInput.getAttribute("placeholder"));
  // ...
```

_assets/javascripts/components/pagination.js
```diff
@@ -83,4 +83,53 @@ CRDS.PageLoader.prototype.activatePage = function(type) {
...
+  const paginationInput = document.getElementById("pagination-current-page");
+  const paginationNextButton = document.getElementById("pagination-next-button");
+  const paginationPrevButton = document.getElementById("pagination-previous-button");
+  const paginationSubmitButton = document.getElementById("pagination-submit-button");
+  const protocol = window.location.protocol;
+  const protocolAndHost = `${protocol}//${host}`
+  const totalPaginatedPages = parseInt(document.getElementById("total_pages").innerText);
+
+  paginationSubmitButton.addEventListener("click", function(e) {
+    if (pageNumber > 1 && pageNumber <= totalPaginatedPages) {
```
dholloran-ample, line 99: It looks like this is the same block of code as starts at line `~110`. Can you abstract it to a function? I'd usually use [the rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) before this but the code here is somewhat complex and rather only maintain it in one place.

dholloran-ample (changes requested):
@mjvezzani 

1. Left inline comments.
1. Please hide the `<<` when on the first page.
1. Please hide the `>>` when on the last page.
1. Please make the input close/equal to the number of characters (Currently it is really large)
1. Please style the buttons I'm ok with any of these options.
    - `<crds-button />` ([link](https://crds-components.netlify.app/?path=/story/ddk-atoms--crds-button)) - It may not work with the script
    - A regular `<button />` with `btn btn-blue`
    - No button background, border, etc. and just display the text

**Note:** This is what I'm seeing it may be related to the inline comment that is causing an error.

screenshot 2021-07-08 at 16 24 53

## Maybe Refactor?

You don't need to refactor and this may help (though it may make it worse 🤷🏻‍♂️ ). Wou could refactor to move all of the `window.location` and page selection logic to something like this. I have not run this code so it may miss some edge case. It is just an example of how I may move some of the repeated code into one place.

It's is personally hard to follow everything that is going on since pagination is complex.

```js
$(document).ready(function () {
  new CRDS.Pagination();

  const totalPaginatedPages = parseInt(
    document.getElementById("total_pages").innerText
  );

  function goToPage(number) {
    const url = `${window.location.protocol}//${window.location.host}`;
    number = isNaN(number) ? 1 : parseInt(number, 10);
    number = number < 1 ? 1 : number;
    number = number > totalPaginatedPages ? totalPaginatedPages : number;

    if (number > 1) {
      window.location.href = `${url}/media/articles/page/${number}/`;
      return;
    }

    return `${protocolAndHost}/media/articles/`;
  }

  const currentPage = parseInt(paginationInput.getAttribute("placeholder"));
  const paginationInput = document.getElementById("pagination-current-page");
  const paginationNextButton = document.getElementById(
    "pagination-next-button"
  );
  const paginationPrevButton = document.getElementById(
    "pagination-previous-button"
  );
  const paginationSubmitButton = document.getElementById(
    "pagination-submit-button"
  );

  paginationSubmitButton.addEventListener(
    "click",
    function () {
      goToPage(paginationInput.value);
    },
    false
  );

  paginationInput.addEventListener(
    "keyup",
    function (e) {
      if (e.key == "Enter") {
        return goToPage(paginationInput.value);
      }
    },
    false
  );

  paginationPrevButton.addEventListener(
    "click",
    function (e) {
      goToPage(currentPage - 1);
    },
    false
  );

  paginationNextButton.addEventListener(
    "click",
    function (e) {
      goToPage(parseInt(currentPage) + 1);
    },
    false
  );
});
```

PR #2368, opened 2021-07-26
US20614 Webfont Visibility

MariaSzubski:
## Problem
[Rally Task](https://rally1.rallydev.com/#/66096747656d/custom/24109743995?detail=%2Fuserstory%2F492689707936)

## Solution
Change `font-display` to `swap` to keep system fonts visible while webfonts load.

### Corresponding Branch
This PR will also need to be merged:
https://github.com/crdschurch/crds-styles/pull/510

## Testing
Run a Lighthouse Performance test on any page. The warning "Ensure text remains visible during webfont load" should not appear in the `Diagnostics` section.

**Note: There are 2 font-faces coming from the `crds-styles` repo.**
`BookWeb` will still appear under this warning until the changes from the linked PR above are merged in.


dholloran-ample (changes requested):
@MariaSzubski

Ok, this is a little weird but the `_font.scss` file is output at build time (took me a while to realize that). So it is overwriting your changes each time the site builds. See Sean's **Option 2** [here](https://www.seancdavis.com/blog/add-typekit-to-jekyll/) for more details on what is going on.

That said we need to set this value in Typekit ([docs](https://helpx.adobe.com/fonts/using/font-display-settings.html)) instead of in `_font.scss`. There should be a Crossroads Typekit account in 1Password.

MariaSzubski:
> @MariaSzubski
> 
> Ok, this is a little weird but the `_font.scss` file is output at build time (took me a while to realize that). So it is overwriting your changes each time the site builds. See Sean's **Option 2** [here](https://www.seancdavis.com/blog/add-typekit-to-jekyll/) for more details on what is going on.
> 
> That said we need to set this value in Typekit ([docs](https://helpx.adobe.com/fonts/using/font-display-settings.html)) instead of in `_font.scss`. There should be a Crossroads Typekit account in 1Password.

@dholloran-ample I've updated the `font-display` setting in Typekit and added a commit with the new auto-generated `_fonts.scss` file.

dholloran-ample (approved):

PR #2426, opened 2021-09-17
add redirect for online registrants

sscarberry:
## Problem
online registrants are being redirected to crds-net and not church online after account activation.

## Solution
redirect based on cookie presence  will allow for said users to be properly redirected.

### Corresponding Branch
/auth-redirect branch see also [this pr](https://github.com/crdschurch/crds-online/pull/342) and t[his pr](https://github.com/crdschurch/crds-components/pull/1450)

## Testing
Testing will require a user to register on Church Online, activate their email and see if redirection is to Church Online Sign in vs. crds-net signin

### DIFF
--- a/redirects.csv
+++ b/redirects.csv
@@ -196,6 +196,6 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /online/getstarted,https://online.crossroads.net/,301!
 /getstarted/thank-you,https://online.crossroads.net/,301!
 /live,https://online.crossroads.net/,301!
-/*,/404.html,404
 /signin,/https://online-int.crossroads.net,301 Cookie=fromOnline
+/*,/404.html,404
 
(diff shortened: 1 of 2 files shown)


redirects.csv
```diff
@@ -197,3 +197,5 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /getstarted/thank-you,https://online.crossroads.net/,301!
 /live,https://online.crossroads.net/,301!
 /*,/404.html,404
+/signin,/https://online-int.crossroads.net,301 Cookie=fromOnline
```
tcmacdonald, line 200: What does `Cookie=fromOnline` do? Also, this file is processed in order so having this rule below the wildcard rule `/*` will prevent this new addition from ever getting evaluated. I'd recommend moving it up one line.

sscarberry: It checks for the presence of the cookie "fromOnilne" and redirects if it is present. Info (Here)[https://docs.netlify.com/routing/redirects/redirect-options/#redirect-by-cookie-presence]

Also good to know or the evaluation order.

dholloran-ample (changes requested):
@sscarberry 
Instead of using the Netlify domains, you should use the `*.crossroads.net` since they are redirects. (My example was a proxy and there is not an actual domain pointing to the Netlify installs)

- `int-crds-online.netlify.app` should be `online-int.crossroads.net`
- `demo-crds-online.netlify.app` should be `online-demo.crossroads.net`
- `crds-online.netlify.app` should be `online.crossroads.net`

Once those are changed then we should be good to go.

dholloran-ample (approved):

PR #2427, opened 2021-09-20
Add check for keys that should be blocked for pagination input

mjvezzani:
## Problem
Pagination input field allowed negative values on /media/articles

## Solution
Add an event listener and checks on keystrokes to prevent the input of anything other than positive numbers in the pagination input field.

### Corresponding Branch
No corresponding branch

## Testing
Navigate to https://6148bc3e9fd22934197af791--crds-net.netlify.app/media/articles/ to see a deploy preview. Try typing letters and characters that aren't numbers in the input field to confirm that they do not appear. Click the up and down arrows in the input field to confirm that you can't navigate to negative numbers.

### DIFF
--- a/_assets/javascripts/components/pagination.js
+++ b/_assets/javascripts/components/pagination.js
@@ -156,8 +156,25 @@ $(document).ready(function() {
     navigateToPageByInput();
   }, false);
 
-  document.getElementById("pagination-current-page").addEventListener("keyup", function(e) {
-    if (e.key == 'Enter' && canNavigate) {
+  document.getElementById("pagination-current-page").addEventListener("keydown", function(e) {
+    const dashKeyCode = 189;
+    const numPadNumber0KeyCode = 96;
+    const numPadMultiplyKeyCode = 106;
+    const number0KeyCode = 48;
+    const colonKeyCode = 58;
+    const backspaceKeyCode = 8;
+    const enterKeyCode = 13;
+
+    if (e.keyCode == dashKeyCode || e.key == "-") {
+      e.preventDefault();
+      return false;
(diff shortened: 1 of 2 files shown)


_assets/javascripts/components/pagination.js
```diff
@@ -156,8 +156,17 @@ $(document).ready(function() {
     navigateToPageByInput();
   }, false);
 
-  document.getElementById("pagination-current-page").addEventListener("keyup", function(e) {
-    if (e.key == 'Enter' && canNavigate) {
+  document.getElementById("pagination-current-page").addEventListener("keydown", function(e) {
+    if (e.keyCode == 189 || e.key == "-") {
```
dholloran-ample, line 160: Could you set the numbers to a variable and then use that? Mainly so we do not have a bunch of magic numbers floating around? (See example below)

```
var something = 189;
var somethingElse = 95;
// ...
if (e.keyCode == something || e.key == "-") {
// ...
```

dholloran-ample (changes requested):
@mjvezzani

- Left 1 inline comment
- Could you also account for `0` as well? Negatives and non-numeric numbers seem to work ok

_assets/javascripts/components/pagination.js
mjvezzani: On it!

mjvezzani: Code has been updated. Here is an updated netlify deploy preview:

 https://6148e137227ea71a0a66ac9c--crds-net.netlify.app/media/articles

mjvezzani:
Here is a new deploy preview url:  https://6148e137227ea71a0a66ac9c--crds-net.netlify.app/media/articles/

dholloran-ample (approved):

PR #2431, opened 2021-09-23
On demand

sscarberry:
## Problem
No way to communicate with members that there is a prior weeks service available for watching during non-service hours.

## Solution
Created CTA to render while it's not service time for viewing the most recent service.

## Testing
Look at designs on screen to verify work against design.

### DIFF
--- a/_assets/javascripts/components/resi-player.js
+++ b/_assets/javascripts/components/resi-player.js
@@ -54,6 +54,11 @@ const refreshPageForServiceStart = (hours, minutes, seconds) => {
   }, timeout);
 };
 
+// For OnDemand Card
+if (isServiceTime() && document.getElementById('ondemand-cta')) {
+  document.getElementById('ondemand-cta').remove();
+} 
+
 if (isServiceTime() && document.getElementById('resi-player')) {
   const resiPlayer = document.getElementById('resi-video-player-container');
   const resiScript = document.createElement('script');
@@ -71,4 +76,5 @@ if (isDayOfTheWeek(0)) {
   refreshPageForServiceStart(8,25,1);
   refreshPageForServiceStart(9,55,1);
   refreshPageForServiceStart(11,40,1);
+  refreshPageForServiceStart(13,30,1);
 }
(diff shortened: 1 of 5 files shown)


_assets/javascripts/components/resi-player.js
```diff
@@ -54,6 +54,11 @@ const refreshPageForServiceStart = (hours, minutes, seconds) => {
   }, timeout);
 };
 
+// For OnDemand Card
+if (isServiceTime() && document.getElementById('ondemand-cta')) {
+  document.getElementById('ondemand-cta').remove();
+} 
+
```
silasbaur, line 61: So I have realized a problem with this solution... There is a 10 minute space between services where this block may show up. This will be an issue if they have not posted the new video until 11.

My suggestion is to create a similar function to the `isServiceTime()` that is specifically for this on demand block. We are going to want it to want this block to not show up between 8:25 and 1:30 on Sundays

_assets/stylesheets/pages/_locations.scss
```diff
@@ -4,7 +4,7 @@
 
   @media (min-width: $screen-sm) {
     padding-bottom: 0;
-    padding-top: 135px;
+    padding-top: 32px;
```
silasbaur, line 7: Why is this reduced `100px`? We should not need to make any big changes to the jumbotron. When this block is not showing we will no longer have the correct spacing for the jumbotron

_includes/locations/_on-demand-card.html
```diff
@@ -0,0 +1,9 @@
+<div id="ondemand-cta" class="ondemand-card">
+  <div class="text-container">
+    <p>MISSED THE LIVESTEAM?</p>
+    <h1>WATCH IT ON DEMAND</h1>
```
silasbaur, line 4: Lets fix the casing on this text and add a class of `text-uppercase`

silasbaur (changes requested):
Looks good so far! We will also need to add this to the Columbus jumbotron. Before this merged we will want to add the contents of the on-demand-card to a content block in the cms but we will handle that right before we merge. Let me know if you have any questions about the comments that I left.

PR #2511, opened 2022-01-06
US21759 Shows Pagination

mjvezzani:
Working implementation of show more episodes on shows, separated by season

## Problem
Currently all episodes display on a show. This creates UX issues of users potentially having to scroll a lot to browse available episodes and has negative performance implications of loading all of the images at once.

## Solution
This PR adds code which implements a load more solution, showing the first 5 episodes of a season and giving the user a "Load More" button that loads an additional 5 episodes. The "Load More" button can be pressed repeatedly until there are no more episodes to reveal, at which point the button goes inactive.

### Corresponding Branch
There are no corresponding branches associated with this work.

## Testing
Given I am a user visiting a show's detail page
When I initially view the list of episodes of a given season
Then I should only see 5 episodes listed
And I should see a "Load More" button

Given I am a user viewing the shows detail page
When I click the "Load More" button at the bottom of the episodes list
Then I should see 5 more episodes load
And all images associated with episodes should appear
And transition from low resolution to high resolution

Given I am a user viewing the shows detail page
And I've loaded more episodes several times
When there are no more episodes to load
Then the "Load More" button should be inactive and can't be clicked


mjvezzani:
[Here is the deployment preview.](https://61d74fba354cdc0064374dc1--int-crds-net.netlify.app/media/collections/garage-bible-study)

silasbaur (changes requested):
Some weird behavior on the second tab. The load more button is a link to a video and below it is a blue container with content that is also clickable


https://user-images.githubusercontent.com/32345656/152389066-4c1bacc7-532f-4423-9134-86ce5a7c19d4.mov

silasbaur (changes requested):
Last thing. Looks like there is a bunch of space at the bottom below the footer when you are on the second tab
Screen Shot 2022-02-03 at 1 02 12 PM

mjvezzani:
@silasbaur Just made changes to get rid of whitespace at bottom of page. I also tested both tabs to ensure that no whitespace, or other position/display issues were happening. I think this is the one!

silasbaur (approved):

PR #2517, opened 2022-01-11
add cypress test to a docker container

lynchl607:
run test in docker container from github action
add automation id for online and live message 
added error handling 
skipped series test
updated the version of cypress from 4.8.0 to 6.5.0
updated version on action from v1 to v2

### DIFF
--- a/cypress/integration/mediaHomePage/landing_page_current_series_spec.js
+++ b/cypress/integration/mediaHomePage/landing_page_current_series_spec.js
@@ -1,9 +1,9 @@
-import { SeriesQueryBuilder} from 'crds-cypress-contentful';
+import { SeriesQueryBuilder } from 'crds-cypress-contentful';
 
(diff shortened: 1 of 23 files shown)


cypress/integration/mediaHomePage/landing_page_current_series_spec.js
```diff
@@ -19,19 +19,19 @@ describe('Testing the Current Series on the Media landing page:', function () {
   });
 
   beforeEach(function () {
-    cy.get('div.media-body').as('featuredSeries')
+    cy.get('body > div:nth-child(9) > section.bg-tan-light.watch-section > div > div:nth-child(2)').as('featuredSeries')
```
silasbaur, line 22: Not a fan of looking for the nth child. Could you add/use a class or an id to the spot that is being tested

silasbaur (changes requested):

silasbaur (approved):

PR #2518, opened 2022-01-11
DE8835 Fix Canonical URL Issues

mjvezzani:
This commit adds a shell script that makes changes to .html files in the
_site directory after the site is built. The changes made are to switch
instances of https://crds-online to https://online.crossroads.net. This
will result in correctly formed canonical urls.

## Problem
Canonical urls that point to Online Church are malformed

## Solution
A shell script has been written that changes instances of https://crds-online to https://online.crossroads.net after the site is done getting built.

### Corresponding Branch
This task has no corresponding branch

## Testing
Visit articles/other media that have crds-online set as their distribution channel in Contentful and confirm that their canonical url (see page source) is set correctly.


tcmacdonald (approved):
Thanks @mjvezzani – this makes sense to me. Can you get this tied to a deployment preview so I can review the build output?

Also, this script is adding a number of new lines of STDOUT to the build output. Considering our other task regarding build logs and reduced verbosity, could we make this either configurable and/or reduce the volume of output by default? Feel free to push back if you foresee challenges or problems with this feedback.

mjvezzani:
Thanks for the feedback. I've reduced the log output for this script down to one line.

tcmacdonald:
Thanks @mjvezzani – can you explain this output to me? 

```
11:23:36 AM: []###### Substitution script complete: Original instances of crds-online / online.crossroads.net: 33 / 2 | New instances of crds-online / online.crossroads.net: 33 / 2
```

Is that saying there were originally 33 instances of `crd-online` and now there's only 2? 

Also, I'm assuming that this PR represents the first step in addressing the issues by doing a string replace in the build output. Do you have another PR for crdschurch/crds-contentful-ui-extensions that addresses the root problem in the CMS?

mjvezzani:
`###### Substitution script complete: Instances of crds-online / online.crossroads.net:       37 /        1 | Instances of crds-online / online.crossroads.net after substitution:        0 /       38`

I’ve changed the nomenclature to clear things up a bit. The first set of numbers indicate how many references to https://crds-online there are vs. how many references to https://online.crossroads.net. The second set show the same comparison after the substitution has been made.

Above we can see that there were originally 37 crds-online references and 1 online.crossroads reference. After substitution we can see 0 and 38, which means that all of the crds-online references contained within the _site directory  were changed.

dholloran-ample (approved):

PR #2522, opened 2022-01-13
Add atrium event redirects

adamtgeorge:
• I tried in contentful but it wouldn’t take

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -7,11 +7,11 @@
 /tags /topics 302!
 /videos/tags /videos 302!
 /weekend-prompts /weekendfollowup 302!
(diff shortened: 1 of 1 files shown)


_redirects
```diff
@@ -8,3 +8,10 @@
 /videos/tags /videos 302!
 /weekend-prompts /weekendfollowup 302!
 /*  /404/index.html 404
+/atriumevents/?site=oakley https://api.crossroads.net/atriumevents/eventlisting/oakley
```
tcmacdonald, line 11: These lines need to come before the catch all `*` rule ☝🏻 otherwise, they will never get processed.

tcmacdonald (changes requested):

tcmacdonald (approved):

PR #2533, opened 2022-01-27
US22090 Update Health Check for CSS and JS

mjvezzani:
## Problem
Health check wasn't catching some problems

## Solution
Added checks for application.css, application.js, and application-deferred.js

### Corresponding Branch
No corresponding branches

## Testing
Confirm Netlify health check runs on a build in Netlify

### DIFF
--- a/_plugins/netlify-plugin-health-check/index.js
+++ b/_plugins/netlify-plugin-health-check/index.js
@@ -9,6 +9,18 @@ module.exports = {
         "_site/_redirects",
         "/*	/404.html	404",
       ]);
+      await run("./bin/health-check.sh", [
+        "_site/assets/application.css",
+        "Twitter, Inc.",
+      ]);
+      await run("./bin/health-check.sh", [
+        "_site/assets/application.js",
+        "jquery.org",
+      ]);
+      await run("./bin/health-check.sh", [
+        "_site/assets/application_deferred.js",
+        "use strict",
+      ]);
     } catch (e) {
       return build.failBuild(e.stdout);
     }


_plugins/netlify-plugin-health-check/index.js
```diff
@@ -9,6 +9,18 @@ module.exports = {
         "_site/_redirects",
         "/*	/404.html	404",
       ]);
+      await run("./bin/health-check.sh", [
+        "_site/assets/application.css",
+        "Twitter, Inc.",
```
silasbaur, line 14: Can you explain why we are checking for "Twitter, Inc."? Seems like we may want it be something like `Bootstrap v3.3.7`

silasbaur: Just looking for a string that better defines the file

silasbaur (changes requested):

_plugins/netlify-plugin-health-check/index.js
mjvezzani: If we look for something like `Bootstrap v3.3.7` and then at some point in the future we upgrade to a newer version and the above text changes, then our build will start failing at no fault of our code, so I think that is a brittle string to check against.

Looking for `Twitter, Inc.` in the application.css file is, in my opinion, a good indicator that the file has both built and exists, as `Twitter, Inc.` is a string that is much less likely to change on a regular basis.

silasbaur: I don't think we will upgrade anytime soon but I'm okay with this argument.

silasbaur (approved):

PR #2578, opened 2022-03-02
Add temporary redirects for 3.20 clicker survey

adamtgeorge:


redirects.csv
```diff
@@ -160,3 +160,13 @@ https://discipleship.crossroads.net/*,https://www.crossroads.net/huddle,301!,mas
 /weekend-prompts,/weekendfollowup,302!
 /atriumevents/ site=:siteId,https://api.crossroads.net/atriumevents/eventlisting/:siteId,302!
 /*,/404.html,404
+/Oakley_qr,https://www.pollev.com/oakley,302!
```
tcmacdonald, line 163: The `/*` rule needs to come last in this file otherwise your redirects will not be observed (e.g. move all of your additions up by one line). 

Also, any particular reason you're using capital letters for the first character of each path? I'd recommend lowercase all the things.

tcmacdonald (changes requested):

redirects.csv
adamtgeorge: Thanks. I'll fix. The upper case came from copy/paste from Google Doc @tcmacdonald

PR #2604, opened 2022-04-28
redirect all countries except US for specicif URL

bennash:
When non-US visitors route to the following page:

https://www.crossroads.net/john-week-4/

... redirect them to this page:

https://www.crossroads.net/content-unavailable/

Solution
Add long redirect rule for all countries except US

Corresponding Branch
Testing
need to use VPN and mimic location from another country


redirects.csv
```diff
@@ -1,3 +1,4 @@
+/may1/*,https://www.crossroads.net/content-unavailable/:splat,"302! Country=AD,AE,AF,AG,AI,AL,AM,AO,AQ,AR,AS,AT,AU,AW,AX,AZ,BA,BB,BD,BE,BF,BG,BH,BI,BJ,BL,BM,BN,BO,BQ,BR,BS,BT,BV,BW,BY,BZ,CA,CC,CD,CF,CG,CH,CI,CK,CL,CM,CN,CO,CR,CU,CV,CW,CX,CY,CZ,DE,DJ,DK,DM,DO,DZ,EC,EE,EG,EH,ER,ES,ET,FI,FJ,FK,FM,FO,FR,GA,GB,GD,GE,GF,GG,GH,GI,GL,GM,GN,GP,GQ,GR,GS,GT,GU,GW,GY,HK,HM,HN,HR,HT,HU,ID,IE,IL,IM,IN,IO,IQ,IR,IS,IT,JE,JM,JO,JP,KE,KG,KH,KI,KM,KN,KP,KR,KW,KY,KZ,LA,LB,LC,LI,LK,LR,LS,LT,LU,LV,LY,MA,MC,MD,ME,MF,MG,MH,MK,ML,MM,MN,MO,MP,MQ,MR,MS,MT,MU,MV,MW,MX,MY,MZ,NA,NC,NE,NF,NG,NI,NL,NO,NP,NR,NU,NZ,OM,PA,PE,PF,PG,PH,PK,PL,PM,PN,PR,PS,PT,PW,PY,QA,RE,RO,RS,RU,RW,SA,SB,SC,SD,SE,SG,SH,SI,SJ,SK,SL,SM,SN,SO,SR,SS,ST,SV,SX,SY,SZ,TC,TD,TF,TG,TH,TJ,TK,TL,TM,TN,TO,TR,TT,TV,TW,TZ,UA,UG,UM,UY,UZ,VA,VC,VE,VG,VI,VN,VU,WF,WS,YE,YT,ZA,ZM,ZW"
```
silasbaur, line 1: Need to get rid of the `"` before the `302!` or is that intentional?
There is also a `"` at the end of the country list and I do not see quotes in the documentation that was provided

silasbaur (approved):
Once the comment is addressed this can be merged

redirects.csv
bennash: Yes those quotes are intentional, as this is a CSV and that string between the quotes is to be parsed as one string. We first tried without quotes then realized that this is a CSV not the _redirects file.

PR #2648, opened 2022-06-28
[Defect] Unified Redirect 

bbbtttiii:
## Problem
Redirect from crds-net to crds-unified was missing a slash

## Solution
Add slash before `:splat` so the redirect works properly

## Testing


jodylane:
confirmed test was failing before this body of work

PR #2682, opened 2022-08-22
8 17 all trips update

adamtgeorge:
• reformat page
• modify overlay-card for trips


adamtgeorge:
@tcmacdonald oh yeah. There are no active trips in prod so it’s blank.  Shall I change it to be development?

tcmacdonald (approved):

PR #2698, opened 2022-09-22
Implement ContentEngaged Event

mjvezzani:
The following changes add requisite code to enable accurate reporting of the ContentEngaged analytics event.

## Task
Add requisite code to enable the new "ContentEngaged" analytics event to get reported.

### Corresponding Branch
There is a corresponing PR in the crds-components repo. https://github.com/crdschurch/crds-components/pull/1747

## Testing
While monitoring the Segment debugger, confirm "ContentEngaged" events get reported while pushing various subscribe buttons, like buttons, share buttons, when videos start, and when videos finish.

### DIFF
--- a/_includes/_article-social-share.html
+++ b/_includes/_article-social-share.html
@@ -4,23 +4,22 @@
 <section class="soft-half-left">
   <div class="row soft-bottom">
     <div class="flex-start flex-align-center">
-      <a href="#" role="button" class="btn btn-blue push-half-right" data-toggle="modal" data-target="#subscribeModalForm"
-      style="width: 105px;">Subscribe</a>
+      <a href="#" id="articleSubscribeButton" data-engaged="subscribed" role="button" class="btn btn-blue push-half-right" data-toggle="modal" data-target="#subscribeModalForm" style="width: 105px;">Subscribe</a>
       <div class="article-detail-social">
         <crds-heart-button theme="button" contentful-id="{{ item.contentful_id }}" contentful-slug= "{{ item.slug }}" contentful-type="{{ item.content_type | upcase }}"></crds-heart-button>
       </div>
       <div class="social-share sidebar">
-        <a href="javascript:shareFacebook()">
+        <a data-engaged="shared" href="javascript:shareFacebook()">
           <svg class="icon" viewBox="0 0 256 256">
             <use xlink:href="/assets/svgs/icons.svg#facebook"></use>
           </svg>
         </a>
-        <a href="javascript:shareTwitter()">
+        <a data-engaged="shared" href="javascript:shareTwitter()">
           <svg class="icon" viewBox="0 0 256 256">
             <use xlink:href="/assets/svgs/icons.svg#twitter"></use>
           </svg>
         </a>
-        <a href="javascript:share();" class="clipboard"
+        <a data-engaged="shared" href="javascript:share();" class="clipboard"
           data-clipboard-text="{{ site.url }}{{ item.url }}"
           data-clipboard-success="{{ site.url }}{{ item.url }} copied to clipboard!">
           <svg class="icon" viewBox="0 0 256 256">
--- a/_includes/_subscribe-buttons.html
+++ b/_includes/_subscribe-buttons.html
@@ -11,7 +11,7 @@
 
         {% if include.youtube %}
         <li>
-          <a href="//www.youtube.com/user/crdschurch" role="button" class="btn btn-cyan btn-subscribe soft-quarter-ends d-flex justify-content-between">
+          <a href="//www.youtube.com/user/crdschurch" role="button" class="btn btn-cyan btn-subscribe soft-quarter-ends d-flex justify-content-between" onclick="dataLayer.push({'event': 'ContentEngaged', 'event_action': 'subscribed'})">
             <div class="d-flex flex-column push-quarter-right align-items-start btn-text">
               <span class="font-size-small push-quarter-bottom">YouTube</span>
               <span class="font-size-smallest white-faded text-uppercase">Subscribe</span>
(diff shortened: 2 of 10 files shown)


_includes/_article-social-share.html
```diff
@@ -4,23 +4,22 @@
...
     <div class="flex-start flex-align-center">
-      <a href="#" role="button" class="btn btn-blue push-half-right" data-toggle="modal" data-target="#subscribeModalForm"
-      style="width: 105px;">Subscribe</a>
+      <a href="#" id="articleSubscribeButton" data-engaged="subscribed" role="button" class="btn btn-blue push-half-right" data-toggle="modal" data-target="#subscribeModalForm" style="width: 105px;">Subscribe</a>
       <div class="article-detail-social">
         <crds-heart-button theme="button" contentful-id="{{ item.contentful_id }}" contentful-slug= "{{ item.slug }}" contentful-type="{{ item.content_type | upcase }}"></crds-heart-button>
       </div>
       <div class="social-share sidebar">
-        <a href="javascript:shareFacebook()">
+        <a data-engaged="shared" href="javascript:shareFacebook()">
```
tcmacdonald, line 12: I'm not seeing any event handlers in this PR that leverage this attribute. Is this just a handler you'll be using in GTM to trigger the event?

_includes/_bitmovin-player.html
```diff
@@ -6,12 +6,13 @@
...
 
-{% assign timestamps = video.timestamps | jsonify %}
+{% assign timestamps = video.timestamps | jsonify | escape %}
 
 {% if recommended_collection.size > 0 %}
   {% assign sorted_collection = recommended_collection | sort: 'date' | reverse %}
-  {% assign json_collection = sorted_collection | jsonify_message: page.date %}
+  {% assign json_collection = sorted_collection | jsonify_message: page.date | escape %}
   <crds-bitmovin-player 
+    recommended-collection='{{json_collection}}'
```
tcmacdonald, line 15: This update seems unrelated. Can you double check that this edit was intentional?

_includes/_subscribe-buttons.html
```diff
@@ -60,7 +60,7 @@ <h2 class="component-header flush-top">Subscribe</h2>
         {% endif %}
         {% if include.content_type.stitcher_url.size > 1 %}
         <li>
-          <a href="{{ include.content_type.stitcher_url }}" role="button" class="btn btn-cyan btn-subscribe soft-quarter-ends d-flex justify-content-between">
+          <a href="{{ include.content_type.stitcher_url }}" role="button" class="btn btn-cyan btn-subscribe soft-quarter-ends d-flex justify-content-between" onclick="({'event': 'media-subscribe-button-pushed', 'event_action': 'subscribed'})">
```
tcmacdonald, line 63: This isn't actually pushing to the `dataLayer`. Should we abstract this logic into a JS file so we don't have 4 instances of the same event handler?

_layouts/article.html
```diff
@@ -21,7 +21,7 @@
           </div>
           <div class="overlay-card-content">
             <p class="overlay-card-category">{{ page.category.title }}</p>
-            <h2 class="overlay-card-title overlay-card-heading">{{ page.title | truncate: 70 }}</h2>
+            <h2 data-media-title="{{page.title}}" data-media-id="{{page.id}}" class="overlay-card-title overlay-card-heading">{{ page.title | truncate: 70 }}</h2>
```
tcmacdonald, line 24: This may not be relevant but the Contentful ID for each page is already exposed in a meta_tag...  

Screen Shot 2022-09-26 at 9 13 24 AM

_layouts/episode.html
```diff
@@ -13,7 +13,7 @@
             <ol class="breadcrumb hard text-white">
               <li><a href="/media/">Media</a></li>
               <li><a href="/media/podcasts">Podcasts</a></li>
-              <li><a href="{{ podcast.url }}">{{ podcast.title }}</a></li>
+              <li><a data-media-title="{{podcast.title}}" data-media-title="{{podcast.id}}" href="{{ podcast.url }}">{{ podcast.title }}</a></li>
```
tcmacdonald, line 16: You have two `data-media-title` attributes here. I'm guessing that is not correct...

tcmacdonald (changes requested):
Left some feedback above @mjvezzani. 

My only other thought is that your naming conventions for this dataLayer thing are very specific to this implementation. Should not the `event` value actually be the event name that you're intending to invoke? For example... 

```
dataLayer.push({'event': 'ContentEngaged', 'event_action': 'subscribed', ...})
```

Seems like this would reduce some of the abstraction both here as well as in GTM and reduce future confusion for others. 

Screen Shot 2022-09-26 at 9 26 43 AM

Please take a look and hit me up w/ questions.

_includes/_article-social-share.html
mjvezzani: @tcmacdonald There is a variable in GTM that is set off this data-engaged attribute. That variable informs a trigger that is part of the ContentEngaged tag.

_includes/_bitmovin-player.html
mjvezzani: @tcmacdonald  So I did all my work, rebased to squash all my commits into a single commit, and then rebased off of master. I didn't notice it before, but for some reason when I pushed to github, two other commits by JerebChase were included in this PR. I don't know why that would be. Any thoughts? What do I need to do to remove his commits from my PR?

tcmacdonald: Got it. This update may not be reflected in `development` and when you pulled down `master` you introduced it. 

As long as this is consistent with what we've got in the `master` branch, I think its safe to leave as-is.

_layouts/episode.html
mjvezzani: Fixed.

PR #2741, opened 2022-11-14
Remove Role=user from Unified Routes

jodylane:
## Problem
We dont want to utilize nf_jwt for tracking authentication on unified, the routes here are currently being proxied to show a user the auth block pages based on nf_jwt.

## Solution
removed role user from unified routes
removed redirects to signin from auth blocked pages because this file has no concept to determine if we are logged in now

## Testing

---
  - https://app.asana.com/0/0/1203331084606721

### DIFF
--- a/redirects.csv
+++ b/redirects.csv
@@ -17,6 +17,7 @@ https://embedint.crossroads.net/*,https://int.crossroads.net/giving/,302!
 https://embeddemo.crossroads.net/*,https://demo.crossroads.net/giving/,302!
 https://embed.crossroads.net/*,https://www.crossroads.net/giving/,302!
 /,${env:CRDS_UNIFIED_DOMAIN},200!
+/home,${env:CRDS_UNIFIED_DOMAIN}home,200!
 /signout,${env:CRDS_UNIFIED_DOMAIN}signout,200!
 /api/*,${env:CRDS_UNIFIED_DOMAIN}api/:splat,200!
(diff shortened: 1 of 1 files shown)


redirects.csv
```diff
@@ -16,26 +16,18 @@ https://forms.crossroads.net/*,https://${env:CRDS_APP_DOMAIN}/forms/:splat,302!
 https://embedint.crossroads.net/*,https://int.crossroads.net/giving/,302!
 https://embeddemo.crossroads.net/*,https://demo.crossroads.net/giving/,302!
 https://embed.crossroads.net/*,https://www.crossroads.net/giving/,302!
-/,${env:CRDS_UNIFIED_DOMAIN}h,200! Role=user
 /,${env:CRDS_UNIFIED_DOMAIN},200!
+/home,${env:CRDS_UNIFIED_DOMAIN}home,200!
```
tcmacdonald, line 20: Changing this to `/home` may not be necessary now that you're updates in crdschurch/crds-unified#327 are coming more into focus. 

If you do update this path, you'll need to update the post registration/login redirect in crdschurch/crds-components to reflect this new path _and_ make sure that there is a clear means by which the user can manually navigate back to the logged in homepage. 

Due to these considerations, I'd consider dropping this change and leaving the logged in homepage where it is.

tcmacdonald (commented):
Left one comment for you @jodylane

tcmacdonald (approved):

PR #2756, opened 2022-12-02
Add canonical link tags to non-distro channel pages

tcmacdonald:
## Problem

The SEO nerds want canonical link tags on every page. Today, we're only rendering these tags when there is [more than one distribution channel](https://github.com/crdschurch/crds-net/blob/master/_includes/_head.html#L43-L49).

## Solution

This PR adds canonicals to all non-distro channel pages.

## Testing

You're looking for the following convention in the page's source... 

Screen Shot 2022-12-02 at 2 10 18 PM

Some examples... 

* https://deploy-preview-2756--int-crds-net.netlify.app/media/
* https://deploy-preview-2756--int-crds-net.netlify.app/media/videos/get-to-know-alli
* https://deploy-preview-2756--int-crds-net.netlify.app/media/videos/get-to-know-alli
* https://deploy-preview-2756--int-crds-net.netlify.app/media/podcasts/spirit-stories
* https://deploy-preview-2756--int-crds-net.netlify.app/spiritual-outfitters/
* https://deploy-preview-2756--int-crds-net.netlify.app/oakley

---
  - https://app.asana.com/0/0/1202832936097549

### DIFF
--- a/_includes/_head.html
+++ b/_includes/_head.html
@@ -47,6 +47,8 @@
     {% else %}
       <link rel="canonical" href="{{ page | get_canonical_host }}">
     {% endif %}
+  {% else %}
+    <link rel="canonical" href="{{ page.url | absolute_url | chomp: 'index.html' }}">
   {% endif %}
 
   {% if page.content_type == "article" %}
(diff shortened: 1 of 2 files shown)


sscarberry (commented):
In the deploy previews I see that it is using the production canonical url:
image

sscarberry:
Looks like that is due to the distribution channel just being set ad the www value in contentful. Seems good to me!

tcmacdonald:
Good call @sscarberry. Thanks for the review.

PR #2827, opened 2023-01-18
Remove GTM container and replace w/ Segment

tcmacdonald:
## Problem

Loading GTM directly via googletagmanager.com is problematic as certain browsers are blocking 3rd party requests. Thus, we want to load GTM via 1st party hostname, so we need to load GTM as a destination of Segment. 

## Solution

Replace GTM snippet with a new Segment source which loads GTM as a destination. This allows GTM to function as it always has but makes it load via CDN proxy analytics.segment.crossroads.net so its not considered a 3rd party request. 

## Related PRs

- https://github.com/crdschurch/crds-unified/pull/369
- https://github.com/crdschurch/crds-net/pull/2763

## Outstanding Tasks

- [ ] Evaluate ENV when loading source

## Testing

...TBD


jacobtellep (approved):

tcmacdonald:
Closed in favor of #2828

PR #2863, opened 2023-02-20
Rock post golive proxy redirects

jodylane:
## Problem

## Solution

### Corresponding Branch

## Testing


netlify.toml
```diff
@@ -156,30 +160,7 @@ package = "./_plugins/netlify-plugin-health-check"
   CRDS_GROUPS_LEGACY_ENDPOINT = "crds-groups-legacy-demo.netlify.app"
   CRDS_INVOICES = "crds-invoices-demo.netlify.app"
   CRDS_VOLUNTEER = "https://crds-volunteer-demo.netlify.app/"
-
```
sscarberry, line 159: This still needs to be here, otherwise this is good

sscarberry (changes requested):

netlify.toml
sscarberry: The whole block that is removed, github ui weird.

sscarberry (approved):

PR #2895, opened 2023-02-28
[INT] Try to solve sporatic errors in transcription_generator

jacobtellep:
## Problem

## Solution

### Corresponding Branch

## Testing

---
  - https://app.asana.com/0/0/1204066110808912

### DIFF
--- a/_plugins/generators/transcription_generator.rb
+++ b/_plugins/generators/transcription_generator.rb
@@ -24,8 +24,10 @@ module Jekyll
           rescue
             message.data['transcription'] = nil
           ensure
-            local_copy_of_remote_file.close
-            local_copy_of_remote_file.unlink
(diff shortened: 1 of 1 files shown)


_plugins/generators/transcription_generator.rb
```diff
@@ -24,8 +24,10 @@ def generate(site)
           rescue
             message.data['transcription'] = nil
           ensure
-            local_copy_of_remote_file.close
-            local_copy_of_remote_file.unlink
+            if local_copy_of_remote_file
```
jacobtellep, line 27: Met with Taylor about this. We agreed that since this code is ran at build time, any risk for a memory leak is acceptable to take.

jodylane (approved):

PR #2926, opened 2023-03-15
prod | add closed caption icon to /watch

jodylane:
## Problem
we want to add closed caption icon to Message Only on /watch header

## Solution
added icon to project
styled the icon to display next to Message only header on /watch

## Testing
the closed caption icon should display next to the message only header


bbbtttiii (approved):
Looks great

PR #3049, opened 2023-07-10
Happenings page redirect

bbbtttiii:
## Problem

## Solution

### Corresponding Branch

## Testing


sscarberry (commented):
I think we would want to make this a 200 type of proxy route instead of a 301. What I am thinking is that if we 301 this route at some point in the future when we don't route everything through this redirects file and it's migrated to crds-unified routes directly we may have issues. We might need to discuss that further.

bbbtttiii:
> I think we would want to make this a 200 type of proxy route instead of a 301. What I am thinking is that if we 301 this route at some point in the future when we don't route everything through this redirects file and it's migrated to crds-unified routes directly we may have issues. We might need to discuss that further.

Yep, I think 200 makes sense in that case.

PR #3058, opened 2023-07-12
testing redirects

jodylane:
## Problem

## Solution

### Corresponding Branch

## Testing


redirects.csv
```diff
@@ -12,6 +12,10 @@ https://embed.crossroads.net/*,${env:CRDS_ROCK_DOMAIN}mygiving/,302!
 /h,/,302!
 /annualreport/2020,https://annual-report.crossroads.net/,301!
 /annualreports,${env:CRDS_UNIFIED_DOMAIN}annualreports,200!
+/watch,/content-unavailable/,"302! Country=IN,NP"
+${env:CRDS_UNIFIED_DOMAIN}watch,/content-unavailable/,"302! Country=IN,NP"
```
sscarberry, line 16: These will not do anything on crds-net. Since it's not really possible that if someone was to navigate to crd-unified.netlify.app/watch that they would ever hit the redirects on this repo.

sscarberry (changes requested):

redirects.csv
sscarberry: When I say these I mean line 16 and 18

sscarberry (approved):

PR #3091, opened 2023-08-07
update source for components

jodylane:
## Problem
Demo by default is looking at the int for crds components we dont utilize that env any longer so thats a problem

## Solution
update default source to look at demo for crds components
update vault variable in demo
update netlify variable

## Testing
easiest way to test this is to open inspector & look at the head tag
in there you should find to link tags with href attributes
those attributes should include this text on demo environment "https://crds-components-demo.netlify.app/dist"
Screen Shot 2023-08-07 at 4 54 10 PM


_plugins/hooks/config.rb
```diff
@@ -1,8 +1,8 @@
 Jekyll::Hooks.register :site, :after_init do |site|
-  if ENV['NETLIFY']
+  if ENV['NETLIFY'] && ENV['CRDS_ENV'] === 'prod'
```
jodylane, line 2: NETLIFY is a "netlify read only variable" that come free from netlify basically is a boolean value to answer is this a build coming hosted from netlify

jacobtellep (approved):

PR #3124, opened 2023-09-13
block nepal & india

jodylane:
## Problem
we need to block nepal & india from viewing messages

## Solution
block specific message from nepal & india on series
block bluteam vision night for nepal & indiad it.*

## Testing
Youll need a vpn to change your location to nepal or india to test this


bbbtttiii (approved):

jodylane:
https://deploy-preview-3124--demo-crds-net.netlify.app/

PR #3135, opened 2023-09-27
Weekend Message - Thursday Updates

bbbtttiii:
## Problem

## Solution

### Corresponding Branch

## Testing


jodylane (approved):
You might also consider ripping the `_resi-player.html` itself out unless its still needed

bbbtttiii:
> You might also consider ripping the `_resi-player.html` itself out unless its still needed

Probably a good call, I figured I would leave it in since they will probably want it for future streaming events, but it'll be easy enough to just add back.

jodylane (approved):

jodylane (approved):

PR #3172, opened 2023-10-23
Add Flex Page Proxies dynamically within crds-net

bbbtttiii:
## Problem
We need the ability to create a proxy for flex pages in Contentful as they get created.

## Solution
Update `redirects.rb` to fetch all flex page entry slugs, then insert them into `redirects.csv` in addition to the existing redirect entries.

## Testing
- View any Contentful flex page in demo and note the slug field for that entry.
- You should be able to visit that url (aka not get a 404 page) by adding the slug to the deploy preview url. This means the redirect was created successfully.
- For example, the slug `/test-flex-page` is viewable at https://deploy-preview-3172--demo-crds-net.netlify.app/test-flex-page

To test the redirects file directly:
- Pull down this branch locally and fetch the latest `demo` entries from Contentful
- Run this command to build `redirects.csv`:
```
ruby bin/contentful-redirects ./redirects.csv
```
- View `redirects.csv` and you should see the flex page redirects added, including drafted pages
- The same should apply if you pull down `prod` entries (but published entries will be added to the redirects file)

---
  - https://app.asana.com/0/0/1205605533387381

### DIFF
--- a/lib/crds/redirects.rb
+++ b/lib/crds/redirects.rb
@@ -30,7 +30,8 @@ class Redirects
 
   def flex_page_redirects
     if ENV['CRDS_ENV'] == 'prod'
-      data = JSON.parse(get_data(@flex_page_options)).dig('items')
+      # data = JSON.parse(get_data(@flex_page_options)).dig('items')
+      data = nil
     else
       data = JSON.parse(get_preview_data(@flex_page_options)).dig('items')
     end
--- a/lib/crds/redirects.rb
+++ b/lib/crds/redirects.rb
@@ -19,7 +19,7 @@ class Redirects
       query: {
         access_token: ENV['CRDS_ENV'] == 'prod' ? ENV['CONTENTFUL_ACCESS_TOKEN'] : ENV['CONTENTFUL_PREVIEW_TOKEN'],
         content_type: 'flexPage',
-        limit: 2000
+        limit: 1000
       }
     }
   end
--- a/lib/crds/redirects.rb
+++ b/lib/crds/redirects.rb
@@ -15,13 +15,28 @@ class Redirects
         limit: 1000
       }
     }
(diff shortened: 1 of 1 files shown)


lib/crds/redirects.rb
```diff
@@ -15,13 +15,28 @@ def initialize
         limit: 1000
       }
     }
+
+    @flex_page_options = {
+      query: {
+        access_token: ENV['CONTENTFUL_ACCESS_TOKEN'],
+        content_type: 'flexPage',
+        limit: 1000
```
jodylane, line 23: 1 draw back here well have to revisit this logic after 1000 entries have been created, but it should work for now

lib/crds/redirects.rb
```diff
@@ -42,16 +57,19 @@ def to_csv!(path = './redirects.csv', debug=true)
       rows.delete_at(n)
     end
 
-    rows.insert(n, *redirects)
+    flex_page_redirects_data = flex_page_redirects
```
jodylane, line 60: we need to add logic to also be able to pull in drafted flex page slugs but they should only be appended to the list if the environment is demo env.

for prod we only want to add published entries to redirects.rb

jodylane (changes requested):

lib/crds/redirects.rb
bbbtttiii: Increased to 2000. Not sure what the actual limitations of this will be but we can always adjust down the road.

jodylane (approved):

lib/crds/redirects.rb
jodylane: contentful limit is 1000

essentially what I think we would have to do here is check the total number of entries & paginate records & batch them together with multiple contentful fetches

PR #3208, opened 2023-11-30
Add livestream toggle to location pages

bbbtttiii:
## Problem
The goal is to use the Livestream model in Contentful to manage livestream events

## Solution
- Fetch livestream JSON from `crds-data` and append to `site.data`
- Add a script that determines whether a livestream should show
- Include the script on all location pages

### Corresponding Branch
https://github.com/crdschurch/crds-unified/pull/1129

## Testing
- Create or modify a livestream entry in Contentful
- Based on the datetime and page selections, the livestream should only show during those times/on those location pages

https://deploy-preview-3208--demo-crds-net.netlify.app/

### DIFF
--- a/_includes/_livestream.html
+++ b/_includes/_livestream.html
@@ -1,4 +1,3 @@
-<!-- Livestream -->
 <script>
   document.addEventListener('DOMContentLoaded', function () {
     let livestreamData = {{ site.data.livestreams | jsonify | strip_newlines
@@ -15,6 +14,7 @@
 
   let isStream = false;
   let playerUrl = '';
+  let entryId = '';
   let currentPageTitle = "{{ page.name }}";
 
   livestreamData.forEach(function (livestream) {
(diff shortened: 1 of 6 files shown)


_includes/_livestream.html
```diff
@@ -0,0 +1,74 @@
...
+
+    let isPageMatch = livestream.pages.some(function (page) {
+      return page === currentPageTitle || page === 'Churchwide (all location pages)';
+    });
+
+    if (isPageMatch && isTimeInRange(startTime, endTime)) {
+      playerUrl = livestream.url || 'https://resi.media/cr-cinc-oh/Manifest.m3u8';
+      isStream = true;
+
+      // Refresh the page one minute after the livestream starts
```
bbbtttiii, line 32: Since these pages are static, force refresh one minute after the start time and one minute after the end time to ensure that the stream time gets reflected.

jacobtellep (approved):

PR #3221, opened 2023-12-18
Uuserway removal christmas-story

sscarberry:
Removes the userway widget from the christmas-story page on crds-net.

### DIFF
--- a/_layouts/no-header-no-footer.html
+++ b/_layouts/no-header-no-footer.html
@@ -22,7 +22,13 @@
   <!-- End of HubSpot Embed Code -->
 
   {% if site.jekyll_env == "production" %}
-  <script src="https://cdn.userway.org/widget.js" data-account="RCwaT3oWqd"></script>
+    {% if page.url == "/christmas-story/" %}
+    {% endif %}
+    {% elseif page.url == "/christmas-story" %}
+    {% endif %}
+    {% else %}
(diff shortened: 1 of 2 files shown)


_layouts/no-header-no-footer.html
```diff
@@ -22,7 +22,13 @@
...
 
   {% if site.jekyll_env == "production" %}
-  <script src="https://cdn.userway.org/widget.js" data-account="RCwaT3oWqd"></script>
+    {% if page.url == "/christmas-story/" %}
+    {% endif %}
+    {% elseif page.url == "/christmas-story" %}
+    {% endif %}
+    {% else %}
+      <script src="https://cdn.userway.org/widget.js" data-account="RCwaT3oWqd"></script>
+    {% endif %}
```
bbbtttiii, line 31: This logic looks like it'll work but I wonder if we could do this instead

```
{% if page.url != "/christmas-story/" and page.url != "/christmas-story" %}
  <script src="https://cdn.userway.org/widget.js" data-account="RCwaT3oWqd"></script>
{% endif %}
```

bbbtttiii (changes requested):

_layouts/no-header-no-footer.html
sscarberry: Yeah that's definitely better

bbbtttiii (approved):

PR #3276, opened 2024-02-14
Our Memories image fix

mjvezzani:
## Problem
Text cards below images on "our-memories" was bumping up after images loaded.

## Solution
Remove data-optimize-image per Taylor's recommendation.

## Testing
visit /our-memories and see if images load correctly and text boxes under images do not bump up on load.


jacobtellep (approved):

jacobtellep:
@mjvezzani looks good. we will want to point this branch at `demo` instead of `main`. Unless I am missing context to why it's going to `main`.

PR #3287, opened 2024-03-07
Added null check for author

yordanthebteam:
## Problem
[Asana Task](https://app.asana.com/0/1201454665996628/1206757340158778)

## Solution

- In brian-tome.html
  - Added nil check for article.author.full_name where iterating articles

## Preview
[/brian-tome](https://deploy-preview-3287--demo-crds-jekyll.netlify.app/brian-tome/)

### DIFF
--- a/_includes/_media-card.html
+++ b/_includes/_media-card.html
@@ -43,6 +43,7 @@
     <h3 class="component-header flush">
       <a href="{{item | media_url }}">{{item.title }}</a>
     </h3>
+    {% if item.author != nil %}
     <p class="card-text soft-quarter-top">
       {% if item.author.id %}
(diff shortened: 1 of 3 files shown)


jodylane (commented):
I think the point of this task was to remove the null authors from this page not to replace them with a hard coded author. 

we've solved this in other areas by filtering null values out of the original data set or adding a conditional around the section that would render the author to check that data is not null. Ill have to check with PM's to see if this would also be an acceptable solution.

jodylane (changes requested):
build still fails when contentful demo environment has a published article with a drafted author.

sscarberry (approved):

sscarberry (approved):

PR #3402, opened 2025-01-21
Import new library conditionally

jacobtellep:
## Problem
The current implementation loads the legacy components library by default. We need a way to conditionally load either the legacy components or the new unified components library based on configuration, to support gradual migration to the new system.

## Solution
- Added configuration support for the unified components endpoint in `_plugins/hooks/config.rb`
- Added a feature flag `USE_UNIFIED_COMPONENTS` to control which library loads
- Modified `_includes/_head.html` to conditionally load either the unified components or legacy components based on the feature flag

### Corresponding Branch
N/A - This is a standalone feature implementation

## Testing
1. Without the feature flag:
   - Set `USE_UNIFIED_COMPONENTS=false` or leave it unset
   - Verify that the legacy components load (multiple script and CSS files)

2. With the feature flag:
   - Set `USE_UNIFIED_COMPONENTS=true`
   - Verify that only the unified components script loads
   - Verify the URL matches the configured `CRDS_UNIFIED_COMPONENTS_ENDPOINT` or falls back to the default

3. Component functionality:
   - Verify that components continue to work as expected regardless of which library is loaded
   - Verify that the `componentsLoaded` event fires correctly in both scenarios


bbbtttiii (changes requested):
I tested this locally on /media. When I set `USE_UNIFIED_COMPONENTS=true`, I am seeing both the legacy and new component scripts load. Not sure if this is the intent based on the testing criteria. 

Screen Shot 2025-01-22 at 8 59 52 AM

The references at the bottom are initiated by a separate `link` tag on the media landing page, but it could be a non-issue since it's just referencing an existing page in demo that also uses legacy components.

bbbtttiii (approved):

PR #3540, opened 2026-04-15
[Demo] Algolia Excluded Deletion

grimicorn:
This will be utilized in crds-unified to exclude these records from deletion in [Algolia - Indexing Scripts + Record Identifiers](https://app.asana.com/1/37330734394864/project/1201454420722044/task/1213913397226267?focus=true).

## Solution
Adds a `deletionExcluded: true` property to all Algolia records that will be used in crdschurch/crds-unified#2998

### Corresponding Branch

## Testing


grimicorn:
Realized that this is not needed since Jekyll is a totally seperate index and we are only deleting records from unified

